Issue #1193, opened 2014-12-10
Add support for Default Values in AdminPresentation Annotation


Jerry77oz:
Allow setting default values on AdminPresentation.

PR #1194, opened 2014-12-10
Add support for DefaultValues in AdminPresentation Annotation.

Jerry77oz:
Added support for allowing default values in the AdminPresentation Annotation. Reference issue is #1193.

When building the form while adding a new entity the we extract the default values and parse them to verify that they are valid based on the `SupportedFieldType`. If the parsing fails we log a WARN and do not use the value.

Also added support for using `today` as a default value on dates which will set the current **now** date.


phillipuniverse:
@Jerry77oz works like a charm nice work. Made a couple of minor changes.

PR #1195, opened 2014-12-11
Fix MergeManager overrides

smmckay:
Properties.keySet() doesn't take default properties into account, so if an override is used then the contents of default.properties is effectively ignored. I suppose the Javadoc could justify the current behavior, but based on the behavior of MergeManager.loadProperties() it looks like the override properties are supposed to overlay default.properties and not replace it entirely.


phillipuniverse:
Confirmed the problem and the fix, thanks for the pull. Kinda weird that it overrides the `keySet()` like that.

Issue #1199, opened 2014-12-18
Refactor product options


charchar:
Product options are currently generated in a few places by looping through Skus, which can become a problem with a large quantity of product option combinations.  We need to instead go through the Product Options and Values themselves to retrieve the appropriate Sku(s).

phillipuniverse:
Closing because #1200.

PR #1200, opened 2014-12-18
#1199 - Refactor the way product options are retrieved

charchar:


phillipuniverse:
@charchar can you remove all private methods/constructors/variables and change them to protected?

Also, are there any changes required on DemoSite for proper operation?

PR #1210, opened 2015-01-05
Changed AND to OR while checking for OrderStatus value if Payment can be applied to Order

vivekjuneja:
Another Payment can be applied to an Order only if the Order Status should not be IN_PROCESS or CSR_OWNED. Hence, changing the AND condition to OR.


phillipuniverse:
Closing based on comment from @elbertbautista

PR #1214, opened 2015-01-08
Removed OnPropertyChanged due to Issue 1197

vivekjuneja:
I changed the **entityListWithUploadForm.html** by removing the **onpropertychanged** attribute. This is because as the pre-existing code was failing for IE 8. 

With this change, I tested on IE 8,9, 10, Chrome (latest - 39.0.x), Firefox, Safari. This change worked for the limited test scenarios for the browsers as mentioned.

Refer:- [Issue 1197] (https://github.com/BroadleafCommerce/BroadleafCommerce/issues/1197)

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/templates/views/entityListWithUploadForm.html
+++ b/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/templates/views/entityListWithUploadForm.html
@@ -4,7 +4,7 @@
     id="assetUploadForm">
     
     <span style="display: block; margin-bottom: 5px; margin-left: 60px;"><i class="icon-camera"></i> Upload Asset</span>
-    <input id="assetUploadFile" type="file" name="file" onpropertychange="$('#assetUploadForm').submit()" value="Upload Asset"/>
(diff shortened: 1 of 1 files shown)


phillipuniverse:
@brundonsmith you've already fixed the issue separately that this was designed to resolve, right?

brundonsmith:
Yeah. It was a while back, but if I remember correctly, I think that adding the onpropertychange attribute was one strategy I tried out to see if it would make things work in IE8, but I don't believe it made any difference. assetSelector.js already wires the event using jQuery, so if @vivekjuneja has tested this in different browsers and it seems to work, I think it's a safe change to make.

vivekjuneja:
@brundonsmith  - You are right. assetSelector.js already wires the event for IE 8 using jQuery. Hence, onpropertychanged attribute was redundant.

phillipuniverse:
Ok, sounds like this can be closed then. Thanks.

vivekjuneja:
@brundonsmith  - Will you be committing the changes to remove the **onpropertychanged** attribute

phillipuniverse:
Sorry about the hasty close, talked to @brundonsmith and now I get what's going on. Merging this in.

Thanks for the pull @vivekjuneja!

Issue #1086, opened 2014-08-18
Add support for a lookup for update collection lookup type


apazzolini:
Sometimes, removing an entity from a normal AdminPresentationCollection should perform an update of the owning entity instead of removing it entirely. Add this third supported type.

Issue #191, opened 2013-07-16
3.0.0-GA Edit SKU shows all product options


tduffey:
When we add a new SKU to a product we only see the options that have been associated with the product. If you then edit the SKU, you see all available product options, including those not associated with the product. We think it should only show the options associated to the product.

PR #1215, opened 2015-01-09
Allow removes in the respond to validation errors

charchar:
This reverts a change made as part of #1086 which seems to prevent CustomPersistentHandlers from persisting remove errors to the Controller via the flash header messages.

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/AdminEntityServiceImpl.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/AdminEntityServiceImpl.java
@@ -20,6 +20,8 @@
 package org.broadleafcommerce.openadmin.server.service;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.broadleafcommerce.common.admin.domain.AdminMainEntity;
 import org.broadleafcommerce.common.dao.GenericEntityDao;
 import org.broadleafcommerce.common.exception.ServiceException;
@@ -51,6 +53,7 @@ import org.broadleafcommerce.openadmin.server.service.persistence.module.BasicPe
 import org.broadleafcommerce.openadmin.web.form.entity.DynamicEntityFormInfo;
 import org.broadleafcommerce.openadmin.web.form.entity.EntityForm;
 import org.broadleafcommerce.openadmin.web.form.entity.Field;
+import org.hibernate.exception.ConstraintViolationException;
 import org.springframework.stereotype.Service;
 import org.springframework.util.Assert;
 
@@ -72,6 +75,8 @@ import javax.persistence.PersistenceContext;
  */
 @Service("blAdminEntityService")
 public class AdminEntityServiceImpl implements AdminEntityService {
+
+    protected static final Log LOG = LogFactory.getLog(AdminEntityServiceImpl.class);
 
     @Resource(name = "blDynamicEntityRemoteService")
     protected DynamicEntityService service;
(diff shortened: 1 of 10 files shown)


phillipuniverse:
@charchar look at what I changed the name of this pull request to and see if it is descriptive.

Also, could you describe what you were trying to do but couldn't?

charchar:
Detailed explanation here:
https://github.com/BroadleafCommerce/QA/issues/191

phillipuniverse:
@charchar can you add some comments inside of the `remove()` method in `AdminEntityServiceImpl`? This method is different than all the others in that class (add(), update()) and this will prevent someone from adding back in the try/catch. I get it now that we need to bubble up exceptions to the controller so it can add the flash attributes.

phillipuniverse:
Copying this below originally from @charchar on the private issue:

> Here's a more detailed explanation of what I've discovered.

> In `AdminEntityServiceImpl` we catch exceptions from CustomPersistenceHandlers like so:

> ```java
try {
    if (transactional) {
        return service.update(pkg);
    } else {
        return service.nonTransactionalUpdate(pkg);
    }
} catch (ValidationException e) {
    return new PersistenceResponse().withEntity(e.getEntity());
}
```

> It looks like this exception catching was to resolve https://github.com/BroadleafCommerce/BroadleafCommerce/issues/312 and https://github.com/BroadleafCommerce/BroadleafCommerce/issues/321 which was initially for `add()` and `update()` methods in `AdminEntityServiceImpl`.  This exception handling was later included in `remove()` as part of https://github.com/BroadleafCommerce/BroadleafCommerce/issues/1086 which was reversed in https://github.com/BroadleafCommerce/BroadleafCommerce/pull/1215 to prevent any `remove()` messaging from getting ignored. 

> This exception handling causes `ValidationExceptions` in the CustomPersistenceHandlers to be ignored and no longer handled in the method `removeEntity()` and added as header flash messages in the Admin. Line 467 of `AdminBasicEntityController` :

> ```java
try {
    service.removeEntity(entityForm, getSectionCustomCriteria(), sectionCrumbs);
} catch (ServiceException e) {
    if (e instanceof ValidationException) {
        // Create a flash attribute for the unsuccessful delete
        FlashMap fm = new FlashMap();
        ...
    }
    ...
    throw e;
}
```

> However, in the `saveEntity()` method of the same class, we handle the error messaging differently. It appears that `ValidationExceptions` from CustomPersistenceHandlers are completely ignored in this case. Line 409:

> ```java
entityFormValidator.validate(entityForm, entity, result);
if (result.hasErrors()) {
    model.addAttribute("headerFlash", "save.unsuccessful");
    model.addAttribute("headerFlashAlert", true);
}
```

> I feel that we should reconsider how we handle this validation and decide on a pattern.

> Also, we should better handle errors from dealing with entity collections.  Line 757 `addCollectionItem()` in class `AdminBasicEntityController`:

> ```java
entityFormValidator.validate(entityForm, savedEntity, result);
if (result.hasErrors()) {
    FieldMetadata md = collectionProperty.getMetadata();
    ppr = PersistencePackageRequest.fromMetadata(md, sectionCrumbs);
    return buildAddCollectionItemModel(request, response, model, id, collectionField, sectionKey, collectionProperty,
            md, ppr, entityForm, savedEntity);
}
```
> This causes a stack trace to show up in a modal in the admin.

phillipuniverse:
@charchar after reviewing your notes (which I am copied directly into this pull request because I think it's generally valuable investigation work) here are some thoughts:

1. The reason that validation is done like that in `add()` and `update()` is because in those methods you actually care about the properties that failed to be saved and in remove() you don't normally. That's my hunch as to why it is like that anyway
2. I feel like making the add(), update() and remove() consistent makes a lot of sense to avoid confusion. I think that you should put `remove()` back the way it was and then handle the validation errors in the same way that is done in `saveEntity()` in `AdminBasicEntityController` where you pass the `Entity` result to the `entityValidatorService`
3. For the flash attributes, does it make sense to go through all of the validation messages that come back in the `Entity` and then try to surface those individually? I'm not sure how hard this would be or if it's valuable at all

phillipuniverse:
I'll finish this up.

phillipuniverse:
Just to close this out, if you want to add validation to a CPH the best thing to do is:

```
Entity entity = persistencePackage.getEntity();
entity.addGlobalValidationError("This has failed validation because x y z");
throw new ValidationException(entity);
```

PR #1221, opened 2015-01-15
Remove Jersey and replace the REST services implementation with Spring MVC

phillipuniverse:
Currently, Broadleaf uses JAXRS via Jersey for our out-of-the-box REST endpoints. There are a few reasons behind this move:

- There is a potential for version collisions with JEE servers that use a different version of JAXRS
- Dubious (at best) continued support for JAXRS endpoints to work with Spring's dependency injection. As an example, the latest version of Jersey (2.x) does not have out of the box support for Spring injection. Some have had mixed luck with trying to make the Jersey 1.x support for Spring work with Jersey 2.x
- Simplifies learning curves for people that are new to Broadleaf and/or new to Java development as everything will now be Spring MVC

That said, this issue should also encapsulate work towards our APIs still being able to function with Jersey, should a user decide to continue to use the old APIs (Broadleaf 3.1 and below) or simply override our implementations.

Some helpful links:

- https://spring.io/blog/2014/12/02/latest-jackson-integration-improvements-in-spring
- http://stackoverflow.com/questions/11492325/post-json-fails-with-415-unsupported-media-type-spring-3-mvc


phillipuniverse:
@brundonsmith I updated my comments on the checkboxes.

brundonsmith:
@phillipuniverse, so, are we still using Jackson even though Spring MVC supports automatic conversion of POJOs to JSON?

phillipuniverse:
@brundonsmith Spring MVC supports automatic conversion of POJOS to JSON via Jackson. We are configuring Jackson to also support XML, which it doesn't do out of the box.

What we are replacing is Jersey, which is an implementation of JAXRS, which is a Java spec for REST APIs. All of the terms sound the same (JAXRS sounds a lot like Jackson) which is why it can be confusing.

brundonsmith:
After testing the Jackson TypeModifier, should I commit the extended wrapper and applicationContext changes or just discard them?

brundonsmith:
Also, when you say "feature docs", are you referring to this page? http://www.broadleafcommerce.com/docs/core/current/broadleaf-concepts/rest/rest-endpoints

phillipuniverse:
@brundonsmith yes and yes.

phillipuniverse:
FYI, @brundonsmith and I verified that the old REST services defaulted to JSON output as well. The only way to change it to XML is to add the header `Accept: application/xml`. Even if you use `Content-Type: application/xml` and leave off the `Accept` header you will still get JSON output.

As confirmed by @ckittrell, the Spring version of the REST services behaves the same way so I think we're good from a backwards-compatibility standpoint.

phillipuniverse:
The only thing left here is a task for @ckittrell to give a more helpful error message if someone leaves off the `Content-Type` header.

PR #356, opened 2015-01-23
Admin session expiration event can lead to loss of data

phillipuniverse:
1. Edit an entity in the admin
2. Spend enough time on the form to warrant a session expiration
3. Submit the form

Result is that the user is taken back to the login page since their session expired. However, after the user logs in, they go back to the edit page but all of new data they filled out is lost.

Potential solutions:
- Show a banner when the session has expired, something like "Your session has expired, please save all data offline before continuing"
- Show a banner warning when the session is < 1min from expiring as well as after it has expired
- Save the POST data on the redirect to the login page and resubmit the original POST after successful login (potential security risk?)


bpolster:
Please attach screenshots of the expire session messaging.

bpolster:
Other questions.     

1.  What would the experience be for a user that opens up 2 tabs for the admin but is only active in one of them?

2. What would the experience be for a user who is actively typing in a RichTextEdit field (e.g. for a blog)?

ncrum:
Previously, when an admin was editing a field or a blog on the admin site for a lengthy period of time, their session would expire before they saved their work. They would then click save, be redirected to the login page, and lose all the work they had just done.

To solve this problem I devised a few safeguards:
+ Listen to key presses and mouse clicks to indicate activity and reset the session timer.
+ Give a visual warning when the session is less than one minute from expiring. 




+ Give a visual warning when the session has expired and notify the admin to save their work offline. 




In answer to the questions Brian posted:

What would the experience be for a user that opens up 2 tabs for the admin but is only active in one of them?

*This was a tough problem to solve since each tab has its own javascript running the page and timer. What I added to solve this was a cookie shared between tabs that recorded the last time the server reset the session. Using this cookie's value, I simply did a check every pingInterval between the timer within the page and what the timer should be according to the cookie. If there was a difference, I simply set the timer in the page back where it should be. So if there is an update on one tab, the other tab will recognize it the next pingInterval and set its timer accordingly.*

What would the experience be for a user who is actively typing in a RichTextEdit field (e.g. for a blog)?

*If they simply type or click the screen their session will be reset (default session time is 15 minutes). They will observe a green visual warning as shown above when the session timer reaches one minute, asking them if they are still there. If the session expires, then they will observe a red visual warning as shown above urging them to save their work and giving them a link to the login page.*

I would appreciate any feedback or suggestions on this issue! @bpolster, @phillipuniverse, let me know what you think.

bpolster:
@ncrum  Looking good so far ...

I'd like to make a few additional changes ... 

Here are a few things to resolve ...
- [ ] Would prefer a Light Box approach instead of the header bar.    
- [ ] This should "grey out" the page so that the background items are no longer clickable. 
- [ ] The message should have a countdown timer indicating when the session will be invalidated (note that it doesn't have to be the same number as the actual session time out on the web server.    For example, my web session could timeout at 30 minutes but I might set the warning at 18 minutes and the invalidate session to 20 minutes.
- [ ] When the timer expires, the system should call a method to invalidate the session and return the user to the login screen with a message (e.g. flash header that contains a your session expired message).
- [ ] I suspect a follow-on problem to be that the login screen itself probably requires the CSRF token which should be ignored during the login flow for a new session.
- [ ] Verify that the cookie is an in-memory cookie

Let me know if you have questions ...

ncrum:
Those are great suggestions, helped me think of a few optimizations too.

However, when the session expires, should I give the user an option so save their work offline, or send them directly to the login page?

bpolster:
We don't really have any offline save capabilities so I think the right answer is to countdown which invalidates the session and returns them to the login page.

There is a security aspect to this as well.     The data might be sensitive and so keeping on the screen for a long period of time is also not ideal.     The approach suggested is what you see with other apps (e.g. banking) that deal with this problem, so I like us following suit with applications that are even more secure.

ncrum:
Sounds good! That's what I figured. I will make that change!

ncrum:
Just to update you guys, I've made a few fixes so far. Still have a few more things to fix, but here is what the popup message looks like. I think I was able to fix that csrf token issue by hitting the "/admin/adminLogout.htm" address just like when you press the logout button. From some testing I did, it seems to have solved things.

![alt text] (http://cl.ly/image/2z002s3i2f1J/Image%202015-01-29%20at%204.20.50%20PM.png "Example")

bpolster:
The dialogue looks good to me.   I haven't tested locally but had a few questions / test cases for you to comment on. 

1.  Confirm that "stay logged in" makes a server request so that session stays alive.
2.  Confirm that "logout" calls "logout"
3.  I think the right answer is that when the session expires, it should also effectively call logout (or other method that invalidates the session on the server) before taking the user to the login page
4.  "/admin" is a web application context and shouldn't be hard-coded

ncrum:
I have met the requirements given so far. However, I will continue to do some more testing to be sure nothing breaks.

- [x] Would prefer a Light Box approach instead of the header bar.    
- [x] This should "grey out" the page so that the background items are no longer clickable. 
- [x] The message should have a countdown timer indicating when the session will be invalidated (note that it doesn't have to be the same number as the actual session time out on the web server.    For example, my web session could timeout at 30 minutes but I might set the warning at 18 minutes and the invalidate session to 20 minutes.
- [x] When the timer expires, the system should call a method to invalidate the session and return the user to the login screen with a message (e.g. flash header that contains a your session expired message).
- [x] I suspect a follow-on problem to be that the login screen itself probably requires the CSRF token which should be ignored during the login flow for a new session.
- [x] Verify that the cookie is an in-memory cookie

![alt text] (http://cl.ly/image/2a3F0a2l0W35/Screen%20Recording%202015-01-30%20at%2003.49%20PM.gif "Session Expiration Message")
*Session Expiration Message with Lightbox Effect*


![alt text 2] (http://cl.ly/image/0f1r3a2q0d0O/Screen%20Recording%202015-01-30%20at%2003.50%20PM.gif)
*Result of Choosing Logout Option*

![alt text 3] (http://cl.ly/image/3K3E173Z0s3g/Screen%20Recording%202015-01-30%20at%2003.52%20PM.gif)
*Result of Letting the Session Expire*

I left the settings for a one minute expiration message. I used 2/3rds of the actual session timeout for the timer. By default, DemoSite is set to 15 minutes. For testing I went into the admin web.xml and set the session-timeout to 2 minutes. I believe for the actual site we should switch the expire message to be around 3 minutes versus the 1 minute I used for testing!

To set the session timeout for the admin DemoSite, add this to the web.xml:
```xml
<session-config>
    <session-timeout>2</session-timeout>
</session-config>
```

@bpolster, @phillipuniverse you might want to check this out.

ncrum:
Brian, I saw your comment as I was submitting this!

In terms of your fourth item, I have a problem. My method of logging out does call "logout." The way I saw that done through other code is by going to adminLogout.htm which is at th:href="@{/adminLogout.htm}" using thymeleaf. 

However, in Javascript I call it like this:
```javascript
BLC.get({
                url : "/admin/adminLogout.htm"
            }, function(data) {
                window.location.replace("/admin?sessionTimeout=true");
            });
```

Is there another way to do that or to process a logout in javascript? (I couldn't even find an adminLogout.htm or where broadleaf processes a logout)

I'd love a more elegant solution, or maybe a way to just send a request to /invalidateSession to process the logout. I did that at first and could call session.invalidate(), but I was having problems with the csrf token.

bpolster:
That approach looks ok to me for item 3 (and potentially item 2 as well).

For item 4, take a look at 'servletContext' variable that is available to keep from hardcoding "/admin".

Example usage in assetSelector.js
```
 $('body').on('listGrid-asset-rowSelected', function(event, link, fields, currentUrl) {
        var json = {
            'assetUrl' : fields['cmsUrlPrefix'] + fields['fullUrl'],
            'adminDisplayAssetUrl' : fields['servletContext'] + fields['cmsUrlPrefix'] + fields['fullUrl']
        }
        
        $('div.asset-selector-container').trigger('assetInfoSelected', json);
        $('textarea.redactor').trigger('assetInfoSelected', json);
    });
```

bpolster:
That approach looks ok to me for item 3 (and potentially item 2 as well).

For item 4, take a look at 'servletContext' variable that is available to keep from hardcoding "/admin".

Example usage in blc-ace.js
```
  BLC.ajax({ 
                url: $hiddenInput.data('fetchurl'),
                noParse: true
            }, function(content) {
                var editor = ace.edit(element);
                editor.setTheme("ace/theme/tomorrow");
                editor.getSession().setMode("ace/mode/" + fileType);
                editor.setValue(content);
                editor.clearSelection();
                
                var prefs = BLCAdmin.ace.getUserPreferences();
                $.each(prefs, function(key, val) {
                    if (key == 'setKeyboardHandler' && val == 'vim') {
                        var s = document.createElement("script");
                        s.type = "text/javascript";
                        s.src = BLC.servletContext + '/js/admin/ace-src-min-noconflict/keybinding-vim.js';
                        $("body").append(s);
                        val = ace.require('ace/keyboard/vim').handler;
                    }
                    editor[key](val);
                });             
```

ncrum:
@bpolster here are my comments on each of those test cases

- [x] Confirm that "stay logged in" makes a server request so that session stays alive.
**"Stay Logged In" calls stayLoggedIn() on click. 
```javascript
stayLoggedIn = function() {
                        $.doTimeout('update');
                        $("#lightbox").fadeOut("slow");
                        activityCount = 0;
                        BLCAdmin.sessionTimer.resetTimer();
                        $.doTimeout('update', BLCAdmin.sessionTimer.getPingInterval(), updateTimer);
                    }
```
The call to resetTimer() does a server request
```javascript
resetTimer : function() {
            BLC.get({
                url : BLC.servletContext+"/sessionTimerInactiveInterval"
            }, function(data) {
                sessionTimeLeft = data.maxInterval * 2 / 3;
                defaultSessionTime = sessionTimeLeft;
                $.cookie("sessionResetTime", data.resetTime);
            });
        }
```
- [x]  Confirm that "logout" calls "logout"

"Logout" has virtually identical functionality to the logout button on the admin page
```html
<a id="expire-button" class="button medium dark radius" th:href="@{/adminLogout.htm}">Logout</a>
...
<a class="small button radius dark" th:href="@{/adminLogout.htm}" th:text="#{header.logout}" />
```

- [x]  I think the right answer is that when the session expires, it should also effectively call logout (or other method that invalidates the session on the server) before taking the user to the login page
- [x]   "/admin" is a web application context and shouldn't be hard-coded
For 3 and 4, here is some code snippets showing when it calls invalidateSession and invalidateSession
```javascript
...
invalidateSession : function() {
            BLC.get({
                url : BLC.servletContext+ "/adminLogout.htm"
            }, function(data) {
                window.location.replace(BLC.servletContext + "/login?sessionTimeout=true");
            });
        }
...
if (BLCAdmin.sessionTimer.getTimeLeft() < BLCAdmin.sessionTimer.getExpireMessageTime()) {

                            if (BLCAdmin.sessionTimer.isExpired()) {
                                $("#lightbox").fadeOut("slow");
                                BLCAdmin.sessionTimer.invalidateSession();
                                return false;
                            }
...
```
That BLC.servletContext was exactly what I was looking for, thank you!

bpolster:
@ncrum This looks great.

@elbertbautista, can you provide a code review of this when you get a chance.    You can delegate to someone else to do testing for you but I'd like you to review the JS code.      You can follow my comments and action items from this thread and see if you have any additional concerns or feedback before we accept this change.

ncrum:
Still need to test for IE, works for Firefox and Chrome

bpolster:
Let's remove the BLC logo from the session expiration popup / messaging.     Some clients rebrand the admin not to include the BLC logo.

ncrum:
Expire message now looks like:

![alt text] (http://cl.ly/image/2M2e423Q3S1z/Image%202015-02-04%20at%203.27.36%20PM.png "Example")

bpolster:
How about "Session expiring ... " in the title bar instead of "Session Expiration Message"?

ncrum:
done, pushing soon

ncrum:
Here it is

![alt text] (http://cl.ly/image/1C2I1w3t0V22/Image%202015-02-04%20at%203.38.30%20PM.png "Example")

ncrum:
Looks like for IE8 it appears as such:
![alt text] (http://cl.ly/image/3O1P3j3g2N0F/Image%202015-02-04%20at%203.51.10%20PM.png "Example")

Functionality is all correct. It looks wonky though and the css lightbox seems broken. Going to look into this.

ncrum:
Example on IE8:

![alt text] (http://cl.ly/image/1e0P2b1p3T3B/Image%202015-02-04%20at%204.55.48%20PM.png "IE8 Example")


Example on Chrome:

![alt text] (http://cl.ly/image/1V1p2R16411U/Image%202015-02-04%20at%205.09.04%20PM.png "Chrome Example")

phillipuniverse:
This draft got reverted, this is being resolved in #1262.

Issue #213, opened 2013-07-25
Remove allPermissions property from AdminUserImpl


phillipuniverse:
AdminUser has the following relationships:

``` java
/** All roles that this user has */
@ManyToMany(fetch = FetchType.LAZY, targetEntity = AdminRoleImpl.class)
@JoinTable(name = "BLC_ADMIN_USER_ROLE_XREF", joinColumns = @JoinColumn(name = "ADMIN_USER_ID", referencedColumnName = "ADMIN_USER_ID"), inverseJoinColumns = @JoinColumn(name = "ADMIN_ROLE_ID", referencedColumnName = "ADMIN_ROLE_ID"))
protected Set<AdminRole> allRoles = new HashSet<AdminRole>();

@ManyToMany(fetch = FetchType.LAZY, targetEntity = AdminPermissionImpl.class)
@JoinTable(name = "BLC_ADMIN_USER_PERMISSION_XREF", joinColumns = @JoinColumn(name = "ADMIN_USER_ID", referencedColumnName = "ADMIN_USER_ID"), inverseJoinColumns = @JoinColumn(name = "ADMIN_PERMISSION_ID", referencedColumnName = "ADMIN_PERMISSION_ID"))
protected Set<AdminPermission> allPermissions = new HashSet<AdminPermission>();
```

an AdminRole also has a list of AdminPermissions. It seems odd to me that we have essentially 2 separate security models: either you can give an admin user roles (which indirectly gives them permissions), or you can skip that relationship and jump straight to permissions.

What about consolidating this into only using roles to give permissions to admin users? Is there some other use case here that you would want to give an admin user permissions by _not_ going through a role?

Looking for input from all but hoping to catch the eye of @jefffischer and @bpolster

PR #1235, opened 2015-01-27
Make saving changes happen through an ajax post in the admin

phillipuniverse:
The idea is to avoid reload the page after a save, and all the fetch overhead that entails. Here's some preliminary tasks around this issue.

1. A save will result in a javascript post of form information to the server
2. The response is consumed by javascript and the page success flash header is displayed
3. Any validation errors are read in the response and the UI is updated to reflect the validation problems in the same way

Related to BroadleafCommerce/QA#213


phillipuniverse:
@ckittrell your IDE is messed up, all tabs should be set to 4 spaces on all of your editors (looking specifically at that Javascript file). If you're using Eclipse, this is where that editor is:

> In Preferences --> General --> Editors --> Text Editors
"Insert spaces for tabs" should be selected
"Show print margin" should be selected, with the column set to 124.
124 is the number of characters visible at a time on GitHub. This isn't an enforced limit (the formatter won't wrap your lines), but you should try to respect 124 where possible
In Preferences --> Web --> Editor
Ensure "Indent using spaces" is selected and the indentation size is set to "4"

phillipuniverse:
The errors object coming back from the save are not being translated in message bundles: 
screen shot 2015-02-04 at 9 16 08 am

Issue #272, opened 2013-08-15
Enhances to RegexPropertyValidator


phillipuniverse:
1. Configurable options on what to do with a null input value
2. Handle incorrectly-configured regex (exception handling)

PR #1262, opened 2015-02-12
Allow graceful session timeouts in the admin so that data is not inadvertently lost

phillipuniverse:
Originally captured and written in #356, reverted because of issues reported in BroadleafCommerce/QA#266 and BroadleafCommerce/QA#272.

One of the problems reported in BroadleafCommerce/QA#266 (`"TypeError: BLCAdmin.sessionTimer is undefined"`) has been resolved so we don't need to keep investigating that one. We should take a closer look at this and try to get this back in shortly after 4.0 GA.

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/controller/entity/AdminBasicOperationsController.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/controller/entity/AdminBasicOperationsController.java
@@ -17,9 +17,11 @@
  * limitations under the License.
  * #L%
  */
+
 package org.broadleafcommerce.openadmin.web.controller.entity;
 
 import org.apache.commons.lang3.StringUtils;
+import org.broadleafcommerce.common.web.JsonResponse;
 import org.broadleafcommerce.openadmin.dto.BasicFieldMetadata;
 import org.broadleafcommerce.openadmin.dto.ClassMetadata;
 import org.broadleafcommerce.openadmin.dto.DynamicResultSet;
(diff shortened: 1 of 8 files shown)


ncrum:
@phillipuniverse, you might want to check this out.

The issue discussed in BroadleafCommerce/QA#272 was likely caused by the following:
+ multiple tabs being opened
+ admin request time > 5 minutes
+ while tabs are loading, session expires in one of the tabs
+ multiple tabs finally finished requests to just get sent back to login screen

I'm not entirely sure how to recreate the slow admin request time, but using the chrome debugger I can kind of emulate it. 

I found that if one tab is held up by a request for instance, then the other tab will not recognize any change in the cookie that holds the last request time. Therefore, this other tab will continue to countdown and inevitably expire the session. This is the main problem I believe.

However, from my testing, I saw that once the slow tab loads, it will change the cookie and the other tab will update its time accordingly. Assuming the session did not already expire.

I made the change to sessionTimeLeft to set it to 999999 every reset before the request just to be safe, but I'm not sure how best to solve this problem. The problem I see is that if the request is taking a long time to go through (e.g. like when you were loading a tab), doesn't that mean the server session has probably not reset? Therefore, the server session could expire before you even get your request back. We could set the cookie temporarily, thereby, resetting the timer for the other tabs before we send the request to reset the session, but the session timer on the server may expire before the request even gets to the server. Then your tabs think the session is still going, yet it actually already expired. I think it might be safer to only reset the timer once the request is received back from the server. 


Also its possible that the reason you didn't see a notification in the first tab is because it automatically redirects to the login page if the session expires, for security reasons. The login page should appear as the image below.

![alt text] (http://cl.ly/image/3d0H0B0i1f0t/Image%202015-02-16%20at%203.34.39%20PM.png)

ncrum:
I also increased the session timeout to be (actual session timeout) - 1 minute instead of 2/3*(actual session timeout).

phillipuniverse:
@ncrum you can recreate the slow timing by setting a breakpoint in the controller for `/sessionTimerInactiveInterval`. The request then would get stuck there.

ncrum:
Doesn't seem to work, but I am running jetty through cmd line instead of eclipse, is that why?

phillipuniverse:
The `site` application listens on remote debug port 8000. The `admin` application listens on remote debug port 8001, regardless of whether or not it's run in Eclipse or not. You should be able to create a remote debug session in Eclipse to port 8001 and using the admin project. For instance:

screen shot 2015-02-16 at 4 13 57 pm

phillipuniverse:
Closing this since I can't change the target branch from multi-tenant-refactor to develop. Now tracking this in #1267.

PR #1273, opened 2015-02-23
Fix copy-paste issue for ProductOptionsWrapper (API)

sshimanski:
NPE happens if productOptionValidationStrategyType is not null, while productOptionValidationType is null.


ktisdell:
@sshimanski we do not accept pull requests on master.  The issue was fixed with the above commit, though. Thanks.

Issue #419, opened 2013-10-02
NPE when sorting products by default category in admin


mihajul:
In the admin products list page if you click the sort icon next to the "Default Category" column, and then click "A-Z", you get the following exception:

```
An error occurred
Unable to fetch results for org.broadleafcommerce.core.catalog.domain.Product

org.broadleafcommerce.common.exception.ServiceException: Unable to fetch results for    org.broadleafcommerce.core.catalog.domain.Product
sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)

java.lang.NullPointerException
org.broadleafcommerce.openadmin.server.service.persistence.module.criteria.CriteriaTranslatorImpl.addRestrictions   (CriteriaTranslatorImpl.java:244)
```

Broadleaf version: 3.0.1

The fix is described in the following diff: 
http://diffchecker.com/hvja0rct
(at line 484 of BasicFieldPersistenceProvider [in current develop branch] move the following condition  !addSearchMappingRequest.getRequestedCto().get(addSearchMappingRequest.getPropertyName()).getFilterValues().isEmpty()  a few lines down, where it's needed.

PR #1289, opened 2015-03-23
Test to ensure that product options are correctly sorted by displayOrder

ncrum:
This pertains to issue BroadleafCommerce/QA#419.
While testing the solution to this issue I found a small oversight where the method was returning productOptions list instead of the sorted list it had just created. 

Here are my steps:
1. Create ProductSpec.groovy to test product options being returned sorted by displayOrder
2. Test failed
3. Changed return value within getProductOptionXrefs() from 'productOptions' to 'sorted'
4. Test succeeded

@phillipuniverse - please take a look and let me know what you think

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/catalog/domain/ProductImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/catalog/domain/ProductImpl.java
@@ -742,7 +742,7 @@ public class ProductImpl implements Product, Status, AdminMainEntity, Locatable,
             }
             
         });
-        return productOptions;
+        return sorted;
     }
 
     @Override
--- a/core/broadleaf-framework/src/test/groovy/org/broadleafcommerce/core/spec/catalog/domain/ProductSpec.groovy
+++ b/core/broadleaf-framework/src/test/groovy/org/broadleafcommerce/core/spec/catalog/domain/ProductSpec.groovy
@@ -90,4 +90,45 @@ class ProductSpec extends Specification {
         output.get(2) == testPox3
         output.get(2).getProductOption() == testPo3
     }
+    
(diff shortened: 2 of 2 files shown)


ncrum:
@phillipuniverse - Knowing spock testing makes testing java a million times easier. I understand why you want people to learn it.

phillipuniverse:
A+ for finding that bug!

Can you add 1 more test: test what happens if you don't set a displayOrder, meaning that it's null? I think that what should happen (according to how I am sorting it) is that it will appear first in the list, followed by all the ones that have specified a `displayOrder`.

ncrum:
done @phillipuniverse

bpolster:
@phillipuniverse - In BLC, we are trying to consistently treat a null as displaying last

ncrum:
From my testing I simply never set a displayOrder for one product option. The result was that option displayed first. The 'displayOrder' instance variable is an Integer. Which defaults it to 'null'. 
This could be fixed to display null last if we use `ObjectUtils.compare(final T c1, final T c2, final boolean nullGreater)` in getProductOptionXrefs() as the following:
```java
public List<ProductOptionXref> getProductOptionXrefs() {
    List<ProductOptionXref> sorted = new ArrayList<ProductOptionXref>(productOptions);
    Collections.sort(sorted, new Comparator<ProductOptionXref>() {

        @Override
        public int compare(ProductOptionXref o1, ProductOptionXref o2) {
            return ObjectUtils.compare(o1.getProductOption().getDisplayOrder(), o2.getProductOption().getDisplayOrder(), true);
        }
            
    });
    return sorted;
}
```

@bpolster should I branch and add an issue for this change?

bpolster:
In this case, tie your commit to this same issue #1289 and reference BroadleafCommerce/QA#419

ncrum:
done

ncrum:
I went ahead and pushed the change to develop since this branch was already merged

Issue #341, opened 2013-09-06
Notify user that admin does not work without JavaScript


bpolster:
If a user disables JavaScript, the admin appears to work but things like Save don't really do anything.    

Instead, we should have some notification that the admin requires JavaScript if it has been disabled.   

Would be nice if the entity actions could target a non-Javascript action (that notifies the user that the admin requires javascript).

PR #1295, opened 2015-03-30
 SiteImpl.java - Site Identifier Type and Site Identifier Value fields are changed as required fields. In order to force site creation with a Site Identifier to avoid exception while preview on site

johnsebastian:
Fix for #341 - Site Identifier Type and Site Identifier Value fields are changed as required fields. In order to force site creation with a Site Identifier to avoid exception while preview on site


phillipuniverse:
@johnsebastian can you make the issue title more descriptive? It looks like you might be using the generated ones from editing files on GitHub; these are almost always insufficient.

johnsebastian:
@phillipuniverse I have updated my comments please verify.

PR #1302, opened 2015-04-02
Added support for new PasswordEncoder in Spring Security 3.2 and deprecate old style

### LINKED ISSUE #1288: Forgot password support uses deprecated PasswordEncoder
`org.springframework.security.authentication.encoding.PasswordEncoder` is deprecated in Spring Security 3.2. Its replacement, `org.springframework.security.crypto.password.PasswordEncoder`, doesn't accept a salt parameter, viz. it exists to support bcrypt. This makes it unusable in the forgot password scenario, because running the token through bcrypt twice will produce two different hashes.

Okay, the reason for salt is that users pick terrible passwords that are easily bruted. This isn't a problem for password reset tokens because users don't pick them. Hashing with SHA-256 or another decent algorithm is sufficient.

P.S. The real problem is that Broadleaf is depending on the deprecated PasswordEncoder. The real solution would be to replace blPasswordEncoder with the new implementation. The issue I'm reporting is just the first I've run into that doesn't have a decent workaround.

pbaggett:
This is designed to fix #1288 in a way that gets us set up to move to the new PasswordEncoder interface.

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/security/service/AdminSecurityServiceImpl.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/security/service/AdminSecurityServiceImpl.java
@@ -378,14 +378,11 @@ public class AdminSecurityServiceImpl implements AdminSecurityService {
         ForgotPasswordSecurityToken fpst = null;
         if (! response.getHasErrors()) {
             token = token.toLowerCase();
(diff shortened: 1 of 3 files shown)


pbaggett:
As apart of this, we were incorrectly using the bean `blPasswordEncoder` injected into the `AdminSecurityService` instead of the correct `blAdminPasswordEncoder`. This should not require any changes from a user perspective but there is a potential that it could. We should notate that in release/migration notes.

PR #1325, opened 2015-04-20
Added logic to report source exception, should rollback routines themsel...

cogitoboy:
Added logic to report source exception, should rollback routines themselves throw an exception.
Noticed that runtime and checked exceptions within the rollback logic would throw an exception out and we would lose the context and nature of the original exception.  This change ensures that the original exception is logged out.


jefffischer:
Closing this pull request. It's pointing to the wrong version/branch of code. This is handled here: https://github.com/BroadleafCommerce/BroadleafCommerce/issues/1333

PR #1118, opened 2015-05-06
Deprecate the customized Broadleaf resource handling in favor of the new resource handlers and transformers in Spring MVC 4.1+

phillipuniverse:
See the docs on the new patterns at http://docs.spring.io/spring/docs/current/spring-framework-reference/html/mvc.html#mvc-config-static-resources. This will end up modifying the `BroadleafHttpRequestHandler` and might also deprecate the `AbstractGeneratedResourceHandler` as now Spring 4.1 has the concept of `ResourceResolver`s that seem to serve the same purpose.

This should be done with some care as it has downstream effects on the Theme (and maybe the AdvancedCMS) module.


phillipuniverse:
Closing, this is superceded by #1351.

PR #1358, opened 2015-05-13
If someone wants to register a CurrencyConversionService there is a exception beeing thrown into CurrencyConversionContext

Pr0TuX:
if someone wants to use this conversion service this would fail, because ``ThreadLocalManager.createThreadLocal(CurrencyConversionService.class)`` will fail because ``CurrencyConversionService`` is an interface and cannot be instatiated


Pr0TuX:
bump :+1:

phillipuniverse:
Sorry about that, gimme a few not sure how this fell out of the radar.

Pr0TuX:
Danke schoen. ;)

phillipuniverse:
Bitte!

Issue #1366, opened 2015-05-20
Hibernate L2 cache isn't updated when adding an item to a media map


smmckay:
Select or upload an image for `skuMedia---primary`, then save and reload the page. If an L2 cache is active, the field will appear to be unpopulated when in fact the appropriate rows are present in `SKU_MEDIA_MAP` etc. AFAICT this is because when `MediaFieldPersistenceProvider` creates a new xref it doesn't actually add it to the containing collection, so Hibernate doesn't invalidate the collection's cache entry. Additionally, it seems like this violates the contract of `FieldPersistenceProvider.populateValue`.

If an L2 cache isn't active then this behavior won't appear, because later sessions will read from `SKU_MEDIA_MAP` and get the correct entities.

PR #1368, opened 2015-05-21
When creating a new Media value, make sure it's added to the parent

smmckay:
Addresses #1366


bpolster:
Thanks for the pull request.    Assigning to @jefffischer to take a closer look.   

@smmckay - you are getting into blc-ninja territory here.   :)

jefffischer:
@smmckay - can you describe the problem you encountered that the pull request resolved, including before and after behavior.

jefffischer:
@smmckay - sorry, ignore. I see that you reference #1366 in your comment.

jefffischer:
The particular solution you propose would interfere with some of the operations of our enterprise module. I understand that you still need a resolution for updating the collection cache state, but we'll need to devise another approach. We'll continue to track this on the #1366 issue you created. I'm closing this pull request for now.

PR #1390, opened 2015-06-05
#1389 - updated robots.txt controller to include setting the site instance

### LINKED ISSUE #1389: robots.txt should be site specific like js/** and css/** files
The sitemap should be site-specific as in this pic:

robots example

ncrum:
Used the `BroadleafContextUtil` method `establishThinRequestContext`  in the `getRobotsFile` function to ensure the robots.txt file uses a site-specific sitemap.

Here are the results of this change on DemoSitePrivate:

example

Fixes #1389


ncrum:
Referencing #1389

phillipuniverse:
@ncrum this looks good, we also need to update DemoSite to move the `robots.txt` URL to go back to being ignored by Spring Security filters.

Issue #663, opened 2014-01-29
change quantity on bundled products does not work


TheoSchumacher:
i'am working with broadleaf demo 3.0.8.GA

when I add a bundled product to the shopping cart, go to 'show the cart' and change the quantity, use update button, the quantity is changed but the total price is not. Going to checkout shows that the price per bundle is now divided by the new quantity instead of multiplying the new quantity with the bundle price to get a new total.

The bundle was of price type 'item sum'.

As I did not change anything to the Demo, I believe this is a bug.

PR #1403, opened 2015-06-09
Add compatibility with SolrCloud Replication

Jerry77oz:
Solr indexing will now use a `Long` instead of `BigDecimal` for `CategoryProductXrefImpl.displayOrder`.

Details can be found here:
https://github.com/BroadleafCommerce/BroadleafCommerce/issues/1402

And here:
https://github.com/BroadleafCommerce/QA/issues/663


phillipuniverse:
@Jerry77oz why convert to a long instead of using a `double`?

Jerry77oz:
@phillipuniverse I was concerned about losing precision. That's why I had opted to multiply it, at which point I could just use a `Long` while keeping the underlying number intact.

Here's some general info I had found on the subject:
http://stackoverflow.com/questions/5749615/losing-precision-converting-from-java-bigdecimal-to-double

It is probably likely that we wouldn't fall into that category but I figured I'd try a "safer" route.

Issue #256, opened 2013-08-07
Create a response object that better details the result of CRUD operations in the PersistenceManager in the admin


jefffischer:
The response should be capable of handling calls for inspect, fetch, update, remove and add. It should contain the standard DynamicResultSet and Entity objects, in addition to a Map for arbitrary additional data that could come back from the call.

PR #1410, opened 2015-06-11
Allow for the primary media for a sku to be removed

### LINKED ISSUE #1228: There is no way to clear out a primary image for a product
If you add a 'Primary Image' to a Product, there is no way to remove that image. You can change it to something else but you cannot actually set it to blank.

This will require modifications to the `MediaFieldPersistenceProvider` to manipulate the defaultSku media map with null values passed in.

cja769:
This fixes #1228 by adding a GUI button that clears the image and value so that "null" is sent to the server. Next the old media object is removed from the database since the newMedia object is null. Lastly the sku-media map entry for its primary image is removed.


bpolster:
This looks like something that should be tested in community

bpolster:
BroadleafCommerce/QA#256

jefffischer:
This is close, but unfortunately it breaks down under the following scenario:

1. Go into the "test" site in demo private
2. Go to Green Ghost and clear the primary image and save
3. Select a new image and save
4. Go to My Changes and revert
5. Go back to the Green Ghost product details screen and you will see the primary image is missing (incorrect)

The problem is that the original clear creates a delete record in BLC_SKU_MEDIA_MAP, and then the subsequent setting of a different image creates an add record in BLC_SKU_MEDIA_MAP. The revert only ends up archiving the add record and the delete record is essentially orphaned.

This enhancement needs to, upon every save, revert any existing changes to primary media before applying a different change to primary media. This guarantees there is only ever one active change for this property in the user sandbox. We do something similar to this for defaultCategory on product. In EnterpriseProductCustomPersistenceHandlerExtensionHandler.manageParentCategoryForUpdate(..), the revertExistingUserEdits(..) call performs a similar function for defaultCategory property. In MediaFieldPersistenceProvider, you need to add an extension point to the MediaFieldPersistenceProviderExtensionHandler interface and implement your similar "revertExistingUserEdits" code in the extension point in WorkflowMediaFieldPersistenceProviderExtensionHandler.

regcole:
@jefffischer for your example in `EnterpriseProductCustomPersistenceHandlerExtensionHandler` ; 
 I noticed that it creates to change records in the following scenario:
1. Go to Product
2. change the Default category and press save.
3. change the default category again and press save.
result: 
image
 Was that the intended behavior to have two change records?

jefffischer:
No - that seems like a bug.

regcole:
@jefffischer nevermind it is not a bug. The code I wrote was causing the duplicate records.

Issue #669, opened 2014-01-30
CMS Spring properties are not available to non-servlet layer beans


jefffischer:
The message source is declared in the servlet context (unlike all the others), resulting in the properties defined being only available to servlet layer beans.

PR #1416, opened 2015-06-12
IsoCountryWrapper needs to be registered bean

regcole:
Pull request to address BroadleafCommerce/BroadleafCommerce#1414 and BroadleafCommerce/QA#669. The `IsoCountryWrapper` should be a registered bean in the framework.


jefffischer:
@regcole Some comments on the repo specific issue (https://github.com/BroadleafCommerce/BroadleafCommerce/issues/1414). Keep in mind, this issue will be used to generate the release notes. Also, in this case, the issue is visible by the community. The title of the issue doesn't really tell someone what's wrong. I would consider changing the title to something like "IsoCountryWrapper is not extensible". Also, for the description, you only have a link to the QA issue, which a community user would not be able to access. There should be a real description here instead. For this issue, it can be short (1 sentence). Simply describe the extensibility issue and the need for the bean to be registered.

jefffischer:
Actually, after looking at the help scout ticket - I guess this may not be an extensibility issue like I assumed. Nonetheless, my statements above hold in that there should be some description of what the problem is.

PR #1417, opened 2015-06-12
Qa 656 collisions

gdiazblc:
Many property comparisons all across BLC use "startsWith" where it might not be the best solution to do so. I created a simple utility function that compares strings based on their respected dot-separated segments.
I suspect there are many other places where this function could be used instead of startsWith, but don't dare to start replacing wholesale.

BTW: I don't know what's up with all those format diffs, our formatting file must have changed. It must be old code; I leave them because I believe the new format more legible.


phillipuniverse:
@gdiazblc can you give this a better description along with referencing the issue that this is supposed to resolve in the description of this pull request?

Issue #955, opened 2014-06-06
Boolean system property should be represented as a radio button or drop down in the admin.


ktisdell:
This is being opened because it was a request from a client.

bpolster:
This is working in 4.0.    Commits above from @cja769 were not merged in.    See notes in BroadleafCommerce/QA#670

Issue #670, opened 2014-01-30
Forgot password functionality no longer works with a random Salt


ktisdell:
In a case where one implements a SaltSource that generates a random salt to store on the password, the functionality to look up the change password token is broken.  In these cases, the salt is often stored with the password (e.g. {hash}:{salt}).  The change password functionality makes the assumption that the salt is the same each time to encrypt the token.  In this case, the salt is stored on the encrypted token rather than being static.  It is also randomly generated, so any additional call to re-encrypt the token and then find it in the database will never work. We need to investigate a work around.

PR #1422, opened 2015-06-15
Boolean system property should be represented as a radio button in the admin.

cja769:
Fixes issue #955 by adding a form builder extension handler to create a new field type, creating a new template that has both a string input and radio button input, and added js functionality to change between inputs depending on the selected item in the system property select list.

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/service/SystemPropertyFormBuilderExtensionHandler.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/service/SystemPropertyFormBuilderExtensionHandler.java
@@ -1,0 +1,42 @@
+package org.broadleafcommerce.openadmin.web.service;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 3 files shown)


phillipuniverse:
Looks like your Javascript editor is using tabs, it needs to be using spaces only. You can fix this by going to Preferences -> Javascript -> Code Style -> Formatter -> Import and importing this file: http://cl.ly/code/0j1e1B2o2y2x/download/broadleaf-javascript.xml

phillipuniverse:
@cja769 for Spring bean annotations (`@Service` `@Repository` `@Controller`, `@Component`) that you add in the framework you need to specify the bean name prefixed with "bl". You currently have `SystemPropertyFormBuilderExtensionHandler` annotated with just `@Component`. Instead, it should be annotated with `@Component("blSystemPropertyFormBuilderExtensionHandler")`.

This is unnecessary in client-specific code (like DemoSite) but necessary in the framework to avoid bean name collisions.

bpolster:
On first review, this seems like an odd way to solve this problem.  

SystemPropertyImpl has a field type that it would seem should be used to set the field type of the property being rendered on the server without any additional client side javascript.

bpolster:
Moving to 4.0.2

bpolster:
Not needed, see notes in BroadleafCommerce/QA#670

PR #1424, opened 2015-06-16
Remove any credit card payment type restrictions on populating billing/shipping/customer information from the order onto a PaymentRequestDTO

### LINKED ISSUE #1423: Allow billing and shipping information to be populated from the order on payment types other than credit card
Currently only credit card payments have the ability to glean things like shipping information, billing information, customer information, etc. This restriction is unnecessary and some third party hosted payments need this information to function properly.

phillipuniverse:
Fixes #1423 by removing any CC type restrictions in `ValidateAndConfirmPaymentsActivity` along with the `ConfirmPaymentsRollbackHandler`. I ran into a situation where a `THIRD_PARTY_HOSTED` payment type needed this info as well and realized the CC type restriction is basically moot.


phillipuniverse:
Reviewed with @elbertbautista.

PR #1428, opened 2015-06-17
#1427 - updated BasicFieldPersistenceProvider to treat null as false when populating values

### LINKED ISSUE #1427: Product sandbox Change Details displaying un-modified fields in admin
Details:
+ Make a change to a product, such as sales price
+ Go to My Changes and promote it
+ View Details for the Promoted Item

Problems:
+ overrideGeneratedUrl showing as being changed from null to blank, it should not show at all.
+ ID showing as being changed from blank to an integer, it should not at all since it is a primary key
+ parent category is showing up as changing from blank to the current value after promotion, it is not being changed at all, so it should not show up either.

Fixed this by changing BasicFieldPersistenceProvider to treat null as false.

ncrum:
Fixes #1427 
Fixes BroadleafCommerce/QA#433.

I made changes to `BasicFieldPersistenceProvider` to properly treat null Boolean values as false by doing a `Boolean.valueOf` on the original value. This fixed the issue where overrideGeneratedUrl would show up on the change details screen.

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/BasicPersistenceModule.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/BasicPersistenceModule.java
@@ -331,12 +331,6 @@ public class BasicPersistenceModule implements PersistenceModule, RecordHelper, 
                     Boolean mutable = metadata.getMutable();
                     Boolean readOnly = metadata.getReadOnly();
 
-                    if (metadata.getFieldType().equals(SupportedFieldType.BOOLEAN)) {
-                        if (value == null) {
-                            value = "false";
-                        }
-                    }
-
                     if ((mutable == null || mutable) && (readOnly == null || !readOnly)) {
                         if (value != null) {
                             handled = false;
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/provider/BasicFieldPersistenceProvider.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/provider/BasicFieldPersistenceProvider.java
@@ -45,8 +45,7 @@ import org.broadleafcommerce.openadmin.server.service.persistence.module.criteri
 import org.broadleafcommerce.openadmin.server.service.persistence.module.criteria.RestrictionType;
 import org.broadleafcommerce.openadmin.server.service.persistence.module.criteria.predicate.IsNotNullPredicateProvider;
 import org.broadleafcommerce.openadmin.server.service.persistence.module.criteria.predicate.IsNullPredicateProvider;
-import org.broadleafcommerce.openadmin.server.service.persistence.module.provider.extension
-        .BasicFieldPersistenceProviderExtensionManager;
+import org.broadleafcommerce.openadmin.server.service.persistence.module.provider.extension.BasicFieldPersistenceProviderExtensionManager;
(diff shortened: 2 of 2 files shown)


jefffischer:
The repo specific issue for this pull request suffers from the same problems I described in your other pull request related to this qa ticket.

jefffischer:
Also, what happens in this code when one or both of the original or new Boolean values are null?

ncrum:
This code basically ensures that the two values are either False or True. So if the origValue was originally null, then it would be converted to False for the comparison, same goes for `populateValueRequest.getRequestValue()`. `Boolean.valueOf` will always return either True or False. The problem with overrideGeneratedValue is that the origValue returned was null, then when it got here, it was finding that null != false and setting it to dirty. I believe the proper behavior is that null == false for Boolean. Does this sound correct?

jefffischer:
I don't believe that null == false for Boolean is always correct. I believe there are cases where Boolean fields are cleared out (i.e. set to null) on purpose. We have a number of Boolean fields in entities where true, false and NULL are valid states.

jefffischer:
For the overrideGeneratedUrl case, I think you have to find out why, even though the field has not been touched in the admin, the value is changing from null to false.

ncrum:
Okay, I will look into that then, perhaps it has to do with ProductImpl defaulting overrideGenerateUrl to false, but I'm not sure yet

ncrum:
This is the line that does it within `createPopulatedInstance`:
```java
if (metadata.getFieldType().equals(SupportedFieldType.BOOLEAN)) {
       if (value == null) {
              value = "false";
       }
}
```

ncrum:
I took out my changes to BasicFieldPersistenceProvider and removed that check in BasicPersistenceModule. The result is that overrideGeneratedUrl does not show up. And we can continue to have null booleans.

jefffischer:
It looks like you removed a block of code from BasicPersistenceModule. I checked the history on that and that code has been there since 1/30/2013. I would suggest that the behavior has been there successfully operating for quite some time. Therefore, my previous statement about nulling out a Boolean value from the admin appears to be incorrect. However, this raises the question, why is overrideGeneratedUrl showing this behavior and not other boolean fields? There's another boolean field on ProductImpl, for example, called canSellWithoutOptions. I would expect to see this problem with all admin managed boolean fields.

ncrum:
I will look into that, because that is a good point, others should show up as well. However, for the other fields, this line doesn't return null, but it does for overrideGeneratedUrl, so I'll look into where those fields within the field manager are being set.
```java
Object origValue = populateValueRequest.getFieldManager().getFieldValue(instance, prop.getName());
```

ncrum:
Turns out overrideGeneratedUrl is a null field within the original entityForm in the controller.

ncrum:
`overrideGeneratedUrl` is null within the database. The other fields that may be null don't show up within the entity form so they don't run into an issue here. I'm not sure what the solution for this should be, we could treat Booleans as only true/false and do my first solution, or allow them to be null and remove that line. Otherwise, we would need to change overrideGeneratedUrl to be stored as non-null in the database. That way when the EntityForm is originally populated it will have false as the value instead of null. What do you think?

jefffischer:
I think I'm going to have to actually run this and review behavior myself more closely. Go ahead and move onto something else for now and I (or someone else) will review this further when we have a chance.

jefffischer:
Btw - the other field example I gave (canSellWithoutOptions ) should show up in the entity form. It just appears on the advanced tab.

ncrum:
Alright, sounds good. Yeah I see it, but for some reason it doesn't show up on the request (looking under Chrome Network Debugger). Perhaps because it isn't set to anything.

jefffischer:
I'm going to close this pull request, since I don't want this code to accidentally get merged later on. I'll reference this discussion in the qa issue for reference.

Issue #1431, opened 2015-06-18
There can be name collisions when multiple assets have the same name


johnsebastian:
This primarily occurs when an asset has been uploaded, deleted, and then a new version of the asset (with the same name) is uploaded.

A) Add ability for images to be physically deleted from the filesystem (when explicitly deleted from the admin)
B) Add caching invalidation across all BLC nodes when images are deleted (this is only relevant if we do not version all uploads)
C) Add versioning on files

PR #1432, opened 2015-06-18
qa 649 asset versioning

johnsebastian:
reference #1431 

This code change is for  versioning the assets while added to the disk as per phase 1, point C requirement for asset management


jefffischer:
There was a problem with your repo specific title/description. I went ahead and fixed - please review for future reference.

johnsebastian:
Thanks Jeff. Will look in future request.

Issue #1440, opened 2015-06-22
403 Forbidden when granted "View Approver Sandbox"


cja769:
There is currently not an out-of-the-box approach to provide view-only permission for the approver sandbox page without exposing additional functionality that should be restricted from an approval view-only user.

phillipuniverse:
Closing, this issue does not belong here. Instead, the approach should modify security permissions for when those specific button actions are added to the listgrid.

PR #1441, opened 2015-06-22
Fixed "View Approver Sandbox" user so that they can view approvals

cja769:
#1440 - The changes made affect how the "View Approver Sandbox" sees the sandbox changes. They no longer see "approve" and "disapprove" buttons so that they don't click them and see a stacktrace. Thymeleaf's spring security dialect was added in order to make it easier to filter out UI elements that lower privileged users don't need to see.

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/resources/applicationContext-servlet-open-admin.xml
+++ b/admin/broadleaf-open-admin-platform/src/main/resources/applicationContext-servlet-open-admin.xml
@@ -125,6 +125,7 @@
         <property name="dialects">
             <set>
                 <ref bean="thymeleafSpringStandardDialect" />
+                <ref bean="thymeleafSpringSecurityDialect" />
                 <ref bean="blAdminDialect" />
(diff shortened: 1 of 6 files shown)


phillipuniverse:
@cja769 the changes that you made to support this change are in the `broadleaf-open-admin-platform` module that does not have an explicit dependency on `broadleaf-framework-web`. Only `broadleaf-admin-module` has this dependency.

However, you added the `thymeleaf-extras-springsecurity3` dependency to `broadleaf-framework-web`and then then used it in `broadleaf-open-admin-platform`. You need to move the dependency into `broadleaf-open-admin-platform` along with this snippet:

```xml
<bean id="thymeleafSpringSecurityDialect" class="org.thymeleaf.extras.springsecurity3.dialect.SpringSecurityDialect"/>
```

phillipuniverse:
Actually, I looked more at what this commit actually does. This is the wrong way to do this. You should instead be modifying that point in which these specific listgrid actions are actually added to the grid.

Closing.

Issue #1442, opened 2015-06-22
Admin presentation setting spill over similarly annotated properties


gdiazblc:
The metadata for an entity property can inadvertently get overridden due to similar property names. There are places in the code where a startsWith comparison is done which may not result in the proper behavior.

If a property name in an admin Entity is a substring of another property's name, then the admin annotation settings tend to spill over the longer-named property. This is due to some rudimentary comparison done with properties, involving startsWith.

The QA issue that originated this is: https://github.com/BroadleafCommerce/QA/issues/656

In that same issue, another BLC issue is referred to as the same as this one:
https://github.com/BroadleafCommerce/BroadleafCommerce/issues/591

I disagree: the root cause might be the same, but this second issue refers to the behavior of metadata overrides. This issue could be reproduced even working with to simple properties the way it's described on QA, without any overrides involved.

Issue #591, opened 2013-12-13
Provide better syntax for targeting admin metadata overrides rather than always using 'startsWith'


phillipuniverse:
Example:

``` java
@AdminPresentationMergeOverrides({
    @AdminPresentationMergeOverride(name = "fulfillmentGroup.total", mergeEntries = 
        @AdminPresentationMergeEntry(propertyType = PropertyType.AdminPresentation.PROMINENT, booleanOverrideValue = true))
})
@AdminPresentationClass(populateToOneFields = PopulateToOneFieldsEnum.TRUE)
@Entity
public class SomeEntity {

    @ManyToOne
    protected FulfillmentGroup fulfillmentGroup;

}
```

With this configuration, not only is the 'total' field in `FulfillmentGroup` prominent, these other fields are now also prominent:
- totalTax
- totalItemTax
- totalFeeTax

This is because those other 3 fields also `startsWith` the `fulfillmentGroup.total` name that I'm specifying in the override. It would be cool if we supported regex when targeting override fields.

The workaround for the above problem is to manually call out those fields that you don't want leaving you with:

``` java
@AdminPresentationMergeOverrides({
    @AdminPresentationMergeOverride(name = "fulfillmentGroup.total", mergeEntries = 
        @AdminPresentationMergeEntry(propertyType = PropertyType.AdminPresentation.PROMINENT, booleanOverrideValue = true)),
    @AdminPresentationMergeOverride(name = "fulfillmentGroup.totalTax", mergeEntries = 
        @AdminPresentationMergeEntry(propertyType = PropertyType.AdminPresentation.PROMINENT, booleanOverrideValue = false)),
    @AdminPresentationMergeOverride(name = "fulfillmentGroup.totalFulfillmentGroupTax", mergeEntries = 
        @AdminPresentationMergeEntry(propertyType = PropertyType.AdminPresentation.PROMINENT, booleanOverrideValue = false)),
    @AdminPresentationMergeOverride(name = "fulfillmentGroup.totalItemTax", mergeEntries = 
        @AdminPresentationMergeEntry(propertyType = PropertyType.AdminPresentation.PROMINENT, booleanOverrideValue = false)),
    @AdminPresentationMergeOverride(name = "fulfillmentGroup.totalFeeTax", mergeEntries = 
        @AdminPresentationMergeEntry(propertyType = PropertyType.AdminPresentation.PROMINENT, booleanOverrideValue = false))
})
[...]

PR #1443, opened 2015-06-22
Admin configuration settings spill over similarly named properties

gdiazblc:
This fix compares properties associated with admin presentation, in a more sophisticated way than just the previous "startsWith". 

As a result, this fixes issue #1442 and issue #591 

The originating QA ticket for all this, was https://github.com/BroadleafCommerce/QA/issues/656

I tested this fix in the following scenarios:

1) **Simple property**
1.1) add a simple property to SkuImpl called "retailPriceEffDate" (starts with the same as an already-existing "retailPrice"
1.2) ensure that "retailPrice" has a REQUIRED annotation argument, while "retailPriceEffDate" doesn't 
1.3) Add accessor methods, expose them on the Sku interface.
1.4) Run admin, try to add a product, verify that retailPrice is required but retailPriceEffDate isn't.

2) **One-to-many property with annotation**
2.1) add a new property to SkuImpl called "retailPriceExtra" (which starts with the same as "retailPrice")
2.2) Mark retailPrice as required via annotation on the property. Ensure that retailPriceExtra is not marked as required
2.3) comment-out the corresponding MergeOverrides annotation on ProductImpl
2.4) Add accessor methods, expose them on the interface
2.5) Run admin, verify that only retailPrice is required

3) **One-to-many property with merge override** 
3.1) remove the require override on the new "retailPriceExtra" annotation
3.2) restore the merge override on ProductImpl for "retailPrice"
3.3) run admin, verify that still only "retailPrice" is required.

Note: There might be other places on the BLC code where a replacement of "startsWith" could be in order.

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/dao/provider/metadata/BasicFieldMetadataProvider.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/dao/provider/metadata/BasicFieldMetadataProvider.java
@@ -17,7 +17,15 @@
  * limitations under the License.
  * #L%
  */
+
 package org.broadleafcommerce.openadmin.server.dao.provider.metadata;
+
+import java.io.Serializable;
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
 
 import org.apache.commons.lang.ArrayUtils;
 import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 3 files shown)


jefffischer:
I'm going to accept this, but only because I don't think this actually addresses #591. #591 complains about property "x.y" also impacting "x.y.z". Looking at the StringUtil code and its javadoc, "x.y" into "x.y.z" would return true, which is similar to the current behavior and would not address #591. This is what we want, though, because there is currently an expectation in the system that setting excluded on a property not only excludes that property, but all properties underneath it. Changing this behavior would produce unexpected results.

I believe #591 is really more about allowing explicit declaration of what scope of influence should be for the annotation. I think the mention of regex is key there, as through a regex expression, one would be able to define whether or not the scope is for the particular property only, or should include all properties under it as well. Until we define that behavior and make the necessary changes, it's dangerous to change what the property scope currently means.

I'm also going to mention in the qa ticket that we should confirm there is no impact to cascaded behavior.

jefffischer:
I manually merged this commit because of some merge conflicts. Closing as a result of the manual merge.

jefffischer:
@gdiazblc Some things to keep an eye out for. 

1) One of your commits did not reference an issue at all - every commit must reference an issue
2) Your other commit reference a QA issue. Commits should reference the repo specific issue.

Issue #581, opened 2013-12-06
Support docs and code-completion for Thymeleaf processors


phillipuniverse:
We can write XML files that correspond to the XSD outlined in the Thymeleaf Eclipse Plugin: https://github.com/thymeleaf/thymeleaf-extras-eclipse-plugin#adding-content-assist-for-your-dialect.

Pretty straightforward. I believe the Eclipse plugin specifically can also look at Javadocs in the Java classes so we might not need to duplicate the docs in the XML file and the Java class (untested though).

PR #1447, opened 2015-06-23
Updated dirty form fields immediately after submission 

### LINKED ISSUE #1446: Dirty states are not being returned immediately after save in the admin
Dirty states are not being updated after submission (border around fields).

cja769:
Fixes #1446 by updating entity controller so that the fields that are dirty are sent in the JSON response and by updating the entityForm.js so that the dirty fields are highlighted based on the JSON response.

Original QA issue BroadleafCommerce/QA#581

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/controller/entity/AdminBasicEntityController.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/controller/entity/AdminBasicEntityController.java
@@ -416,8 +416,36 @@ public class AdminBasicEntityController extends AdminAbstractController {
         if (result.hasErrors()) {
             populateJsonValidationErrors(entityForm, result, json);
         }
+        List<String> dirtyList = buildDirtyList(pathVars, request, id);
+        if (!dirtyList.isEmpty()) {
+            json.with("dirty", dirtyList);
+        }
 
         return json.done();
+    }
+    
+    public List<String> buildDirtyList(Map<String, String> pathVars, HttpServletRequest request, String id) {
+        List<String> dirtyList = new ArrayList<>();
+        String sectionKey = getSectionKey(pathVars);
+        String sectionClassName = getClassNameForSection(sectionKey);
+        List<SectionCrumb> sectionCrumbs = getSectionCrumbs(request, sectionKey, id);
+        PersistencePackageRequest ppr = getSectionPersistencePackageRequest(sectionClassName, sectionCrumbs, pathVars);
+        ClassMetadata cmd = null;
+        Entity entity = null;
+        try {
+            cmd = service.getClassMetadata(ppr).getDynamicResultSet().getClassMetaData();
+            entity = service.getRecord(ppr, id, cmd, false).getDynamicResultSet().getRecords()[0];
+
+        } catch (ServiceException e) {
+            return null;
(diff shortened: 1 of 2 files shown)


phillipuniverse:
The modifications to `entityForm.js` is not because `BLCAdmin.workflow.initializeDirtyStates(fieldLabel.closest("form"));` does not exist in outside of the Enterprise/changesets module being present. What this actually means is that there will be a JS error on every single form submission which is no bueno.

What should happen instead is that if you look in that same JS code that you modified, this gets invoked at the end:

```javascript
BLCAdmin.runPostFormSubmitHandlers($form, data);
```

So really, all you need to do is move the JS code in `entityForm.js` out of the framework and into a `postFormSubmitHandler`.

The controller part in `AdminBasicEntityController` is ok but not quite right. I've commented on code changes here inline. Also, you're doing a separate lookup on the `Entity` to determine dirty state. Does this actually work? If you're not running Enterprise, does it actually give back the correct dirty state?

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/controller/entity/AdminBasicEntityController.java
```diff
@@ -416,10 +416,38 @@ public String saveEntityJson(HttpServletRequest request, HttpServletResponse res
         if (result.hasErrors()) {
             populateJsonValidationErrors(entityForm, result, json);
         }
+        List<String> dirtyList = buildDirtyList(pathVars, request, id);
+        if (!dirtyList.isEmpty()) {
```
phillipuniverse: Based on `buildDirtyList()` below, if you catch a `ServiceException` you just return null. If that happens, this line of code will throw a `NullPointerException` because `dirtyList` is null.

You should get into the habit of using `CollectionUtils` as much as possible. In this case, this code should transform to:

```java
if (CollectionUtils.isNotEmpty(dirtyList))
```

Which gives you a couple of advantages:

1. Readability is 100x better
2. It automatically protects against NPEs. If `dirtyList` is null, then that method will return `false` (it treats `null` as empty)

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/controller/entity/AdminBasicEntityController.java
```diff
@@ -416,10 +416,38 @@ public String saveEntityJson(HttpServletRequest request, HttpServletResponse res
...
     
+    public List<String> buildDirtyList(Map<String, String> pathVars, HttpServletRequest request, String id) {
+        List<String> dirtyList = new ArrayList<>();
+        String sectionKey = getSectionKey(pathVars);
+        String sectionClassName = getClassNameForSection(sectionKey);
+        List<SectionCrumb> sectionCrumbs = getSectionCrumbs(request, sectionKey, id);
+        PersistencePackageRequest ppr = getSectionPersistencePackageRequest(sectionClassName, sectionCrumbs, pathVars);
+        ClassMetadata cmd = null;
+        Entity entity = null;
+        try {
```
phillipuniverse: I wouldn't put this in a `try/catch` block. Instead, you should mark the method with `throws ServiceException` and make the caller handle it (in this case, `saveEntityJson()` which is already marked as `throws Exception`.

PR #1450, opened 2015-06-23
Bundles can return a 404 in BLC multi-tenant setup

### LINKED ISSUE #1448: In a multi-tenant environment where tenants share bundle resources, a 404 can be cached for bundle resources
When a bundle resource is accessed for the first time, the system tries to create it.    If multiple threads are all trying to use the bundle, the threads block until the first one creates it.    

The code in `ResourceBundlingService` has a race condition that can occur when multiple threads try and create the same bundle from different `site` contexts.    This can only occur in a Multi-Tenant installation but the race condition is in core Broadleaf.

bpolster:
Resolves #1448 

After creating a bundle, add the bundle resource to the createdBundles map.    Do not add a filename to the map if it wasn't created in the current thread.


phillipuniverse:
Looks good to me.

PR #1455, opened 2015-06-24
Qa 685 admin entity not found

### LINKED ISSUE #1454: 500 Error in Admin when entity not found
In the case of an entity not found error, we want to throw a new specific exception ("EntityNotFoundException") that can be targeted in our exception handler.

Once we have the correct exception identified we can redirect the user to an appropriate page, instead of throwing a stack trace.

jfleschler:
Fixes #1454
When a user tries to access an entity that does not exist, they will be redirected up one level to the main listgrid view for the entity type.

This issue was found by switching sandboxes while on an entity that does not exist in the target sandbox.


phillipuniverse:
I'm going to make these updates because I have some bandwidth but here are some comments:

1. **all** if statements should have curly braces. The use that you have there is not as bad as it could be but we should be standardizing on that
2. When you see things like `entities == null || entities.length != 1`, you should take the opportunity to use `ArrayUtils` or `CollectionUtils` for verifying empty and non-null at the same time.

Issue #1464, opened 2015-06-25
Find a way to display validation errors on a Translation modal form


gdiazblc:
The controller guiding Translation crud operations creates a simplified version of an EntityForm, unsuitable for the structure that the ErrorProcessor normally expects in order to display errors and continue with the flow. 

I have to figure out a way to achieve displaying the validation errors generated by the persistence handler, and avoid the flow of the code to go back to the translation grid regardless.

The originating QA issue for this is https://github.com/BroadleafCommerce/QA/issues/622

Issue #622, opened 2014-01-07
Editing an entity with MatchesFieldValidator in use will throw an NPE


jefffischer:
1. Create a user
2. Edit a field (e.g. change the phone number) and save

The edit will fail and the system will complain about NPE in the MatchesFieldValidator. This only effects 3.1.0 and is not a problem in 3.0.x.

PR #1470, opened 2015-06-26
prevent unwanted translation permutations

gdiazblc:
This PR would avoid entity field translations with unwanted, possibly overlapping results. 
The way it achieves it, is by disabling the internationalization submenu, disabling the locale combo on edition, and adding error handling to the ad-hoc Translations form  (which had none at all).

The originating QA issue was https://github.com/BroadleafCommerce/QA/issues/622

This closes issue #1464, which was created to encompass BLC work on this.

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/handler/TranslationCustomPersistenceHandler.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/handler/TranslationCustomPersistenceHandler.java
@@ -1,0 +1,98 @@
+/*
+ * #%L
+ * BroadleafCommerce Open Admin Platform
+ * %%
+ * Copyright (C) 2009 - 2014 Broadleaf Commerce
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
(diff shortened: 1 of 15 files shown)


jefffischer:
At first glance, this seemed possibly more complex than it needed to be. However, it looks like the translation stuff is "off the range" on how it renders and interacts, so I would imagine to support the requirements a lot of controller and js customization was required.

Can you explain why the special "preventSubmit" functionality is needed? In general, our validation works such that a form is submitted, and if there are validation errors, the persist process exits and the errors are reported to the user in the web browser. The user can submit again, and if the same problems are still there, he/she will see the same error messages again.

Also, I don't see where you have fulfilled the International Messages requirement. I just see the handling for Translations.

gdiazblc:
@jefffischer I believe  the "prevent submit" is necessary because the errorProcessor expects a structure of tabs and field groups that, for Translations, is just not there. If I didn't do anything different when errors are found, the current Javascript code just resubmits always to the grid.

gdiazblc:
Apparently, the International Messages page was already protected against repeated permutations of key/locale. Screenshot attached:

screen shot 2015-07-22 at 5 04 18 pm

gdiazblc:
One more comment on this: long-term, perhaps the proper solution for translations is not this contribution (which is a little jerry-rigged), but a rethinking of the translation's EntityForm and controller, to bring them more in line with the rest of the entities. In that way, all the error processing and the javascript flow in the edition screen would not need much special treatment.

PR #1478, opened 2015-06-30
Fixes BroadleafCommerce/QA#691

ktisdell:
Fixes BroadleafCommerce/QA#691Ensures that filter queries that are
facet queries use an "OR" operator for values.  Also tags and excludes
facet fields from filtering to get proper value counts.


bpolster:
Spoke with Kelly on this.    Change was made for a client and is not demonstrable in the Demo code.    Looking to change the way facets work in the demo site in another ticket to better use the updated approach.   Accepting this into 4.0.1 to solve an immediate client need.

Issue #1479, opened 2015-06-30
nullable=false field on Phone object having unwanted effects


gdiazblc:
On the Address (CustomerAddress) object that is used when inserting a new address in the multi-shipping form, the primary phone should not be required.

There is a many-to-one relationship between Address and primaryPhone.
PrimaryPhone has a property phoneNumber, with a "nullable" attribute set to false, which semantically makes sense, since why would we persist a phone otherwise.

However, setting this nullable="false" seems to have some undesired "contagion" effect into Phone instances altogether.  There is a many-to-one relationship between Address and phone, and, although a primary phone is not required by default, that setting in the phone implementation causes Hibernate to reject attempts to persist an address, unless it has a primaryPhone specified.

I dislike touching this property, and tried many other things around this (different settings for the @ManyToOne and @JoinColumn annotations for the mapping of primaryPhone in AddressImpl), but this nullable=false setting seems to trump them all.

PR #1480, opened 2015-06-30
nullable property on primaryPhone

gdiazblc:
This pull request reverts the nullable = false setting for phoneNumber in the Phone implementation, because it seems to have some undesirable effects when trying to make  many-to-one mappings to this entity optional.

It closes issue #1479 , which I just created to this effect.

The originating QA issue was https://github.com/BroadleafCommerce/QA/issues/695


bpolster:
A few issues ...
1.  We can't accept pull requests that have a series of unrelated commits - closing.      This is occurring frequently with @gdiazblc commits
2.  It doesn't make sense to me that a Phone object would not require a phone number so I am hesitant that the approach taken here is the correct one

phillipuniverse:
I see the problem. @gdiazblc you branched from 4.1 and opened a pull request against 4.0.

gdiazblc:
I was thinking about this: my assumption might not be the whole truth. If that property alone were the cause of the Hibernate exception, then no Phone anywhere could be omitted, ever.
And a proof that it doesn't happen, is in the "Add New Address" form itself, which has also a "secondary phone", but allows omitting it.

Then, there has to be another explanation, related to the DynamicForm, and the binding it does to the Phone class. At first I discarded this, because I made the experiment of totally removing the corresponding tag from the HTML code, and the error persisted.  But perhaps it is a binding problem on lower layers, the DynamicForm against Spring, or a DynamicField being created anyway despite being non-required. I am not sure.

What I can say about this, is that, until we figure it out, the other related fix (non-required dynamic fields on the SQL) will be insufficient for the form not to consider this as required.

Issue #1460, opened 2015-06-24
Add pricelist-related capabilities to the Search Facet Range grid


gdiazblc:
Add the pricelist name to said grid, plus custom settings for its default sort order(s).

The QA issue  originating this one is https://github.com/BroadleafCommerce/QA/issues/621

Also related, is the PriceList issue https://github.com/BroadleafCommerce/PriceList/issues/35 , describing a small annotation enhancement necessary for this.

Issue #1492, opened 2015-07-02
Incorporate an "order" member to FilterAndSortCriteria, FilterMapping, use it for sorting


gdiazblc:
The map contained in CriteriaTransferObject should be able to keep track of the order in which FilterAndSortCriteria, or, if not the CTO itself, there should be a way to provide an order i which sorts will be applied, relative to each other. 

Originally solved via a LinkedHashMap on the CTO, a more solid solution would be in order, something the many different places from which FilterAndSortCriterias could be added/generated.

The idea is incorporating an "order" field to the F&SC (and, on to the FilterMapping), and ensure that the order indicated is taken into account, before building a query.

PR #1493, opened 2015-07-02
additions to SearchFacetGrid fields, and "order" member to Filter&Sort and FilterMappings 

gdiazblc:
This pull request is made against 4.1 instead of 4.0, as it should be, because I had mistakenly checked out against 4.1, originally. Otherwise, it would contain commits made by other people.

It replaces a former pull request that contained only the first commit (issue #1460)

What this pull request does, is offering a way to keep track in which order FilterAndSortCriteria are added to the CriteriaTransferObject, and ensure that such order is taken into account when building the query.

It closes issues #1460 and #1492


gdiazblc:
I will try to redo these commits to the proper milestone.

PR #1494, opened 2015-07-03
additions to SearchFacetGrid fields, and "order" member to Filter&Sort and FilterMappings

gonzalo-roberto-diaz:
This pull request replaces a former pull request that contained only the first commit (issue #1460)
It also supersedes a previous pull request of the same title, but made agains the wrong release (4.1).

What this pull request does, is offering a way to keep track in which order FilterAndSortCriteria are added to the CriteriaTransferObject, and ensure that such order is taken into account when building the query.

It closes issues #1460 and #1492

### DIFF
--- a/admin/broadleaf-admin-module/src/main/resources/bl-admin-applicationContext.xml
+++ b/admin/broadleaf-admin-module/src/main/resources/bl-admin-applicationContext.xml
@@ -45,6 +45,7 @@
                 <ref bean="blSkuCustomPersistenceHandler" />
                 <ref bean="blISOCountryPersistenceHandler"/>
                 <ref bean="blSkuBundleItemCustomPersistenceHandler" />
+                <ref bean="blSearchFacetRangeCustomPersistenceHandler" />
             </list>
         </property>
     </bean>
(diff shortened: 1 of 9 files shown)


jefffischer:
@gdiazblc For future reference, the branch name for this pull request is incorrect. It should be formatted according to "qa-[qa issue number]-[short description]"

jefffischer:
I don't see where the order on FilterMapping is actually ever used. I think rather than sorting the FASC, you should be sorting the final FilterMapping, which is actually used to construct the OrderBy clause on the query. Look a CriteriaTranslatorImpl.addRestrictions(..). It makes more sense to sort the filter mapping collection here before processing. This covers any cases where filter mappings have been explicitly added via API in a custom persistence handler, or the like.

gdiazblc:
@jefffischer, I am sorry that the formatting change makes difficult to see what the actual changes are. I have a hard time myself :) reviewing my own additions.

The map sorting is done my the utility method `processCtoMap(cto);` which is called on line 837 of commit f58c07ee792958f3605fb513ada92093f97530a1

`getFilterMappings()` is not the place where you suggested to put it, but I believe it to be a good place, as it seems that all maps used will have to go through that method. Am I wrong?

jefffischer:
You're sorting the FASC instances. Instead, you should be sorting the FilterMapping instances. getFilterMappings covers the sorting of FASC instances, which results in FilterMapping instances being returned that are in the correct order. However, you are not accounting for the additionalFilterMappings field on the CriteriaTransferObject that is used in a number of CPH implementations for explicit FilterMapping usage. The best place to sort is to sort the FilterMappings before they are used to build the query.

gdiazblc:
Got it. Besides, it would then start to make sense to have passed those "order" values on to the FilterMappings.
Correcting ...

jefffischer:
I assume you closed this to create a new pull request. No big deal, but you probably could have continued to use this same pull request.

gdiazblc:
Yes, I also created a new, properly named qa-... branch

Issue #1496, opened 2015-07-06
Make the catalog name translatable


gdiazblc:
Self-explanatory.
The generating QA issue for this was: https://github.com/BroadleafCommerce/QA/issues/707

Issue #707, opened 2014-02-13
  Fulfillment groups inconsistency with product bundles


TheoSchumacher:
testing the Demo site, I got the following strange behaviour

I have created a product bundle with 2 product: sudden death sauce and sweet death sauce.

If I add this product bundle to an empty order, I have one (bundle) item in the cart with 2 products as items for this bundle.
Looking to the fulfillment groups, I get 2! fulfillment groups, one for each item in the bundle.

If I add first a standard product to the cart, e.g. green ghost sauce, and then adding the bundle, the items view is still ok, that is 1 item for green ghost sauce and 1 item for the bundle. Looking to the fullfillment groups, I get only one fulfillment group with all three (discrete) items. This is what I would have expected.

So, when adding a bundle top an empty cart, creation of fulfillment groups seams to go wrong.

PR #1497, opened 2015-07-06
Make the catalog name translatable

gdiazblc:
This pull request would add the translation capability to the Catalog' s "name" field.
It closes issue #1496 

The originating QA issue was https://github.com/BroadleafCommerce/QA/issues/707, which itself derived from another QA issue requiring this field and another one.


gdiazblc:
Creating another pull request with a branch properly prefixed with "qa-".

Issue #621, opened 2014-01-07
Don't used required validation on offer target criteria, as a blank value is a valid case in 3.1


jefffischer:
Fixed in commit f75019ff308ccb14994f556ed30755017d2a8580

PR #1501, opened 2015-07-07
order field in FilterMapping

gdiazblc:
This pull request improves a previous one  https://github.com/BroadleafCommerce/BroadleafCommerce/pull/1494
by moving the comparison of sort orders later on in the flow of code (see Jeff's comments regarding where to perform the sorting by "order" field).

The original issue addressed, is the necessity of multiple sort orders on a custom persistence handlers's fetch action, and being able to specify them in a certain order.

Addresses QA issue: https://github.com/BroadleafCommerce/QA/issues/621
Closes BLC issues: #1460 and #1492


jefffischer:
My other comment is that I think there will be occasions in which the order field on either FASC or FilterMapping is null. I think this would cause an NPE in your comparator. Your comparator should be updated to account for null and position them last.

gdiazblc:
I am making the order field an object type Integer from the beginning, deliberately leaving it at null of not initialized, passing it on to FilterMapping "as is", and considering it the biggest value in the comparison, if null.

Issue #516, opened 2013-11-09
Discount Type should be marked as 'required' for Offers


phillipuniverse:
There has to be some context for how to apply the value.

PR #1505, opened 2015-07-10
Make the `requiredFlag` field visible by default in admin application.

regcole:
This pull request address BroadleafCommerce/QA#516

The commit does not necessarily "fix" the issue of when a user tries to preview or save a page with data not compatible with a HTML template. The goal of this commit it to provide better "out-of-box" transparency on how to create a PageType that enforces some structural rules that Pages of the PageType must follow.

### DIFF
--- a/admin/broadleaf-contentmanagement-module/src/main/java/org/broadleafcommerce/cms/field/domain/FieldDefinitionImpl.java
+++ b/admin/broadleaf-contentmanagement-module/src/main/java/org/broadleafcommerce/cms/field/domain/FieldDefinitionImpl.java
@@ -114,7 +114,8 @@ public class FieldDefinitionImpl implements FieldDefinition, ProfileEntity {
     protected Boolean textAreaFlag = false;
     
     @Column(name = "REQUIRED_FLAG")
-    protected Boolean requiredFlag = false;
+    @AdminPresentation(friendlyName = "FieldDefinitionImpl_requiredFlag", order = 4000,gridOrder = 4000,defaultValue = "false")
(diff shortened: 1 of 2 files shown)


jefffischer:
Can we get steps on how to create the error in the first place. Also, can you be more clear about how your commit addresses the issue. I'm fine if your fix somehow prevents the problem in the first place, it's just unclear to me right now how it does that.

regcole:
Steps to create the error: 
   1. Navigate to Content >> Pages >>  Featured Armageddon. 
   2. Delete the Product and save
  3. click "preview on site"
Current result: You will get a Thymeleaf error since the template gets a NPE when trying to look for the "product" pageField. 

Ideal Result: The ideal situation would be that the FieldDefinition for "product" would have been marked the field as `isRequired`. This would have prevented the user from being able to save and preview a page that would cause template errors. 

How the commit address the issue: The commit turns on the visibility of the `isRequired` field when creating a new `FieldDefinition`. This way;the user can create a new "Product" `FieldDefinition` for the "FeaturedProuct" `PageType` and mark it as 'isRequired'. Once the field is mark as required, content pages do not get saved or previewed that could potentially cause template errors.

phillipuniverse:
@regcole your reasoning etc and implementation looks ok, but a few comments:

1. Use spaces after commas or put all the annotation attributes on a new line
2. It seems like there is an unintended side effect as you changed the `protected Boolean requiredFlag = false;` to just `protected Boolean requiredFlag;`. I see that you added the `defaultValue` attribute to `@AdminPresentation` which is good, but you should keep the definition as it was.

Issue #386, opened 2013-09-19
Add an activity to send the order confirmation email


phillipuniverse:
This **must** be done in 3.1 and cannot be backported to 3.0.x since this is a change to the blCheckoutWorkflow.

Also relates to BroadleafCommerce/DemoSite#16

PR #1506, opened 2015-07-10
overlapping metadata overrides 

### LINKED ISSUE #811: Using mo:override does not merge configurations with the same ceiling entities
Consider this case:

Module 1:

```xml
<mo:override id="blMetadataOverrides">
        <mo:overrideItem ceilingEntity="org.broadleafcommerce.core.catalog.domain.Product">
             <mo:field name="defaultSku.quantityAvailable">
                <mo:property name="excluded" value="true" />
            </mo:field>
        </mo:overrideItem>
</mo:override>
```

Custom site:

```xml
<mo:override id="blMetadataOverrides">
        <mo:overrideItem ceilingEntity="org.broadleafcommerce.core.catalog.domain.Product">
            <mo:field name="url">
                <mo:property name="excluded" value="false" />
                <mo:property name="visibility" value="HIDDEN_ALL" />
            </mo:field>
        </mo:overrideItem>
</mo:override>
```

The final merged XML presente to Spring is:

```xml
<mo:override id="blMetadataOverrides">
        <mo:overrideItem ceilingEntity="org.broadleafcommerce.core.catalog.domain.Product">
            <mo:field name="url">
                <mo:property name="excluded" value="false" />
                <mo:property name="visibility" value="HIDDEN_ALL" />
            </mo:field>
        </mo:overrideItem>
</mo:override>
```

As the ceilingEntity is the same for both `mo:overrideItem`s, I would instead inspect this final config:

```xml
<mo:override id="blMetadataOverrides">
        <mo:overrideItem ceilingEntity="org.broadleafcommerce.core.catalog.domain.Product">
            <mo:field name="defaultSku.quantityAvailable">
                <mo:property name="excluded" value="true" />
            </mo:field>
            <mo:field name="url">
                <mo:property name="excluded" value="false" />
                <mo:property name="visibility" value="HIDDEN_ALL" />
            </mo:field>
        </mo:overrideItem>
</mo:override>
```

Same goes for `configurationKey` on the `mo:overrideItem`.

gdiazblc:
This pull request creates some special treatment for those metadata override entries in the set of our application context, having overlapping "ceilingEntities".  

If a given ceilingEntity was specified more than once along the different metadata overrides in different application context files, the existing code just overwrote the overrideItem entity with the last value processed. This meant that only the fields of the last processed item were present in the resulting, merged applicationContext.

With this pull request, overlapping metadataOverrides "merge" their field items, and no information is lost. 

Additionally, some refactoring of the NodeReplaceInsert functions is done, extracting potentially reusable code to a utility class.

This fix is very solid: not only merges the overrides as desired, but also;
* if the same field is mistakenly specified more than once, it doesn't duplicate it in the merged result
* it is fault-tolerant with nodes not specifying attributes, or not giving the expected attribute name ("ceilingEntity" for overrideItem, "name" for field).
* the admin UI reacts as expected

The originating QA issue is   https://github.com/BroadleafCommerce/QA/issues/386
It also fixes #811, the issue underlying that QA entry.

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/extensibility/context/merge/handlers/MetadataOverrideNodeReplaceInsert.java
+++ b/common/src/main/java/org/broadleafcommerce/common/extensibility/context/merge/handlers/MetadataOverrideNodeReplaceInsert.java
@@ -17,8 +17,10 @@
  * limitations under the License.
  * #L%
  */
+
 package org.broadleafcommerce.common.extensibility.context.merge.handlers;
 
+import org.apache.log4j.Logger;
 import org.w3c.dom.Node;
 
 import java.util.List;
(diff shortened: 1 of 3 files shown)


phillipuniverse:
Quick note: in your commit message, use this format: `<org>/<repo>#<issuenum>`.

In this case, rather than putting "QA issue #386" you should instead say `BroadleafCommerce/QA#386`. Based on what you wrote, Github parses that as issue 386 from the current repository which is not actually what you want.

Issue #717, opened 2014-02-14
Improve ResourceMinificationService reliability


apazzolini:
Includes leveraging `System.getProperty("line.separator")` to get line separators as well as adding the ability to minify individual resources to assist in troubleshooting. There was also an input stream being used incorrectly.

PR #1507, opened 2015-07-13
Qa 717 widget master data

Jerry77oz:
Reference for QA issue can be found here: https://github.com/BroadleafCommerce/QA/issues/717

This pull request adds support for displaying master group data to a widget. The intention is to not incur a domain change. The approach taken is a naming convention, to have `Fields` and `FieldGroups` use **master** in their name and have that be the driving factor in showing a `dynamicForm` for `Master Content Data`.

As the `dynamicForms` are getting generated we check for master fields groups, and if found we show/hide the respective group. This allows users to have one instance of `Master Content Data` that will be filled out once and then have multiple iterations of `Content Data`.

The totality of this effort is split into this pull request done on `Core`, and another one found in `AdvancedCMS`.


jefffischer:
@Jerry77oz For future reference, your commits need to reference the QA issue

Issue #738, opened 2014-02-25
Roll the Basic Inventory module functionality into the framework


phillipuniverse:
Targeting 3.1.1-GA this adds back inventory concepts to the framework. The workflow activities and necessary Sku property are not hooked up by default as that indicates a breaking change. Instructions for how to add are in the 3.1.1 migration notes (all optional operations).

The activities and property will be included and enabled out of the box in 3.2 and should be included in any migration notes from 3.1 to 3.2.

PR #1512, opened 2015-07-17
Provide mechanism to skip validation for MVEL strings in FieldData

bpolster:
Provide support for MVEL expressions used by FieldData (rules) to utilize expressions that are not just field.property style expressions.    This commit allows a FieldData to set a "skipValidation" property which will avoid the strict check that is intended to catch typos in custom FieldData objects.

This pull-request also provides a small enhancement to RequestDTO which will allow the property map on BroadleafRequestContext to be used.

BroadleafCommerce/QA#738

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/rulebuilder/dto/FieldData.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/rulebuilder/dto/FieldData.java
@@ -17,6 +17,7 @@
  * limitations under the License.
  * #L%
  */
+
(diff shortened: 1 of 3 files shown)


phillipuniverse:
@bpolster looks like your IDE is misconfigured to use tabs instead of spaces and also wrong import organization (should be org, com, java, javax). In `AbstractRuleBuilderFieldService` you're also tabbed over too far.

Formatting docs on 2 from: https://github.com/BroadleafCommerce/docs-internal/blob/master/onboarding/Onboarding-Process.md#setting-up-eclipse

bpolster:
thanks, I'll double check both of those things.    I actually did this in a branch and then did an eclipse compare and move changes dialogue that might be the reason.   I also upgraded eclipse recently so let me check all of those things to make sure it is correct going forward.

PR #1513, opened 2015-07-17
Add ability to remove the primary media from a Product.

regcole:
- Addresses BroadleafCommerce/BroadleafCommerce#1228
- Addresses BroadleafCommerce/QA#256

<h2>Problem</h2>
There is no way to remove the "primary media" once it has been set for a `Product`

<h2> Knowledge use to arrive to solution </h2>
- The "primary media" is a `SkuMediaXref`
- Logic already exist to remove/alter a `SkuMediaXref` linked to a product

<h2> Solution </h2>
Add a button for the media field that sets the value of the input field and replaces the current image/thumbnail with a placeholder image.


phillipuniverse:
Closing this as this does not actually remove the `SkuMediaXrefImpl` entry, it just sets the URL to an empty string. #1410 has the right idea there are just a few bugs in the implementation.

Issue #470, opened 2013-10-29
Where to create new custom files in broadeaf ?


deepak241990:
Hello Mr.Philip Universe
I had a question which is how to add a controller,services and dao. I saw your discussion which is with ankit in broadleaf forum the link for your reference is http://forum.broadleafcommerce.org/viewtopic.php?f=14&t=1889 regarding that you have shared a docs for that may i know the link please.because it will be helpful for me for the further use.

In previous comment the url is http://docs.broadleafcommerce.org/dev/A ... orial.html, 
but it is not openning.

regards
deepak

PR #1518, opened 2015-07-20
Use the `StaticAssetPathService` to properly resolve images

regcole:
Addresses BroadleafCommerce/QA#470

<h3>Steps to reproduce problem:</h3>
1. set `asset.server.url.prefix=[some cdn] ` in  common-shared properties:
2. in admin application Catalog >> Products >> "Sudden Death sauce"
3. add an image in the description box. save,promote,and approve change
4. navigate to the product on the site
Result: If you check the `url` src of the image that is part of the description, you will see that the image was served from the application and not the CDN. 

<h3>Solution:</h3>
Use a Thymeleaf CMS expression parser that replaces cmsstatic with CDN urls. 

<h3>How to Test fix:</h3>
1. In product.html file; around line 36 where template tries to display the description on the page, 
 insert this line `  <div th:utext="${#cms.fixUrl('__*{longDescription}__')}" id="description"></div>`
2. set `asset.server.url.prefix=[some cdn] ` in  common-shared properties:
3. in admin application Catalog >> Products >> "Sudden Death sauce"
4. add an image in the description box. save,promote,and approve change
5. navigate to the product on the site

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/web/expression/AssetURLVariableExpression.java
+++ b/common/src/main/java/org/broadleafcommerce/common/web/expression/AssetURLVariableExpression.java
@@ -1,0 +1,63 @@
+/*
+ * #%L
+ * BroadleafCommerce Common Libraries
+ * %%
+ * Copyright (C) 2009 - 2015 Broadleaf Commerce
+ * %%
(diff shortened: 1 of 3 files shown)


regcole:
updated comments about solution to reflect changes made by commit @f47ec0a

PR #1520, opened 2015-07-21
Creates a DirectCopyTransformer version that conditionally executes based on a Spring property

### LINKED ISSUE #1515: Create a DirectCopyClassTransformer that can use a Spring property to decide whether it should execute
It is sometimes advantageous to conditionally perform class transformation of domain classes based on a property. This would be primarily used during a patch release lifecycle where there is an expectation of no alteration to database schema (i.e. the release should be "drop-in"). However, it is still advantageous for specific implementations, that are willing to absorb the schema adjustments immediately, to consume the fix. They may still do so by setting a property defined as part of the issue resolution, which would engage the class transformation and subsequent schema changes.

jefffischer:
Fixes #1515
BroadleafCommerce/QA#454

It is sometimes advantageous to conditionally perform class transformation of domain classes based on a property. This would be primarily used during a patch release lifecycle where there is an expectation of no alteration to database schema (i.e. the release should be "drop-in"). However, it is still advantageous for specific implementations, that are willing to absorb the schema adjustments immediately, to consume the fix. They may still do so by setting a property defined as part of the issue resolution, which would engage the class transformation and subsequent schema changes.

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/extensibility/jpa/copy/OptionalDirectCopyClassTransformer.java
+++ b/common/src/main/java/org/broadleafcommerce/common/extensibility/jpa/copy/OptionalDirectCopyClassTransformer.java
@@ -1,0 +1,87 @@
+/*
+ * #%L
+ * BroadleafCommerce Common Libraries
+ * %%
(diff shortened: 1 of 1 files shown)


bpolster:
Merging in this utility class.

PR #1522, opened 2015-07-22
QA#764 - Add property to not return the default value for a translated field if that value is not in the same language.

bpolster:
Instead, return blank.   The new property to drive this behavior is named "returnBlankTranslationForNotDefaultLocale".

This ticket also provides an overridable method in TranslationServiceImpl to exclude some fields from this logic.     In particular, the CMS lookup fields (which start with pageTemplate) needed to be excluded for the home page to render in heat-clinic.     This is because all page template fields currently go through translation provider (even fields like ID) which are not even translatable.

Relates to BroadleafCommerce/QA#764

## Testing 
- Set  "returnBlankTranslationForNotDefaultLocale" to true.
- Update the DB so that a language does not have an out of box field (for example, removing the french longDescription entries provides a good test.
- View the home page in english and french.
- With the property set, you should not see any longDescriptions for the products for which the 

For example, see the first 2 products below which had the longDescription removed from the translation table.    Note that with the above property set to "false" (the default), the english translation would show.   (see screenshot 2).

screen shot 2015-07-22 at 1 33 14 am

Also, of note is that the TranslationServiceImpl provides a way to exclude some fields from this behavior.    This can be done by adding items to the `blTranslationExceptionProperties` which by default is set to the following.

```
    <bean id="blTranslationExceptionProperties" class="org.springframework.beans.factory.config.ListFactoryBean" >
        <property name="sourceList">
            <list>
                <value>pageTemplate.*</value>
            </list>
        </property>
    </bean>    
```

> The list is a list of regular expressions.   If they match the property being translated then that property will default to the direct value regardless of the `returnBlankTranslationForNotDefaultLocale` property.

> The `pageTemplate.*` regular expression is required for anyone using foreign key lookups from the BLC CMS and therefore is included by default.   Other items can be added to the `blTranslationExceptionProperties` List component.

> Also of note, the methods added to TranslationServiceImpl provide for the ability for implementations to override if the out of box solution is not sufficient.     Prior to this commit, there was no ability to override as business logic was contained in the static method of `DynamicTranslationProvider`.

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/i18n/service/DynamicTranslationProvider.java
+++ b/common/src/main/java/org/broadleafcommerce/common/i18n/service/DynamicTranslationProvider.java
@@ -50,6 +50,8 @@ public class DynamicTranslationProvider {
             
             if (StringUtils.isNotBlank(translatedValue)) {
                 valueToReturn = translatedValue;
+            } else {
+                valueToReturn = translationService.getDefaultTranslationValue(obj, field, locale, defaultValue);
             }
         }
             
--- a/common/src/main/java/org/broadleafcommerce/common/i18n/service/TranslationService.java
+++ b/common/src/main/java/org/broadleafcommerce/common/i18n/service/TranslationService.java
@@ -130,4 +130,23 @@ public interface TranslationService {
      */
     Cache getCache();
 
+    /**
+     * Intended for use with the {@link DynamicTranslationProvider} to determine the default value when a 
+     * translation was not provided.
+     * 
+     * The default implementation of this method relies on a system property "returnBlankTranslationForNotDefaultLocale". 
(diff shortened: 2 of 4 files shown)


jefffischer:
Would it make sense to change:

if (returnBlankTranslationForNotDefaultLocale) {
   if (localeMatchesDefaultLocale(locale)) {
   } else if (!propertyInDefaultLocaleExceptionList(entity, property)) {
      return "";
   }
}

to 

if (returnBlankTranslationForNotDefaultLocale && !localeMatchesDefaultLocale(locale) && !propertyInDefaultLocaleExceptionList(entity, property)) {
                return "";
}

gdiazblc:
By the way, the French translation should read "... le plus piquant poivre du monde"

PR #1531, opened 2015-08-03
Failed to Lookup another Page Template in page

kunner:
When we choose another page template by lookup, it returned always old template.
look 130 ~ 139 line.
Even template has requested by user's lookup, getFieldGroups always return page.getPageTemplate.
Thats why we got always same template.

It seems need to change two if statements order.


phillipuniverse:
Forum discussion: http://forum.broadleafcommerce.org/viewtopic.php?f=15&t=4453

PR #1532, opened 2015-08-03
Missing parent's pages from ajax dynamicform call 

kunner:
When we click lookup button in Page to choose Page Template and select a row in modal window, we encounter error message like below.

"Unable to perform inspect for entity: org.broadleafcommerce.cms.page.domain.PageTemplate"

it is cause of request parameter for parent's page is missing when ajax dynamicform call.

It seems to need to some change code.

related to below.
open_admin_style.js.admin.components.listGrid.js


phillipuniverse:
Forum discussion: http://forum.broadleafcommerce.org/viewtopic.php?f=15&t=4453

PR #1556, opened 2016-03-04
Added reply-to support in email

squidpickles:
It would be useful for us to be able to customize the reply-to address when sending emails. This patch allows that to happen.


squidpickles:
Let me do this properly... missed that part in `CONTRIBUTING.md`.

PR #1568, opened 2016-09-20
Mark.du patch 1 

khanmark:
Fixes issue with hard coded map key friendly name.

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/service/FormBuilderServiceImpl.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/service/FormBuilderServiceImpl.java
@@ -1196,7 +1196,7 @@ public class FormBuilderServiceImpl implements FormBuilderService {
         } else {
             keyField = new Field().withName("key")
                                 .withFieldType(SupportedFieldType.STRING.toString())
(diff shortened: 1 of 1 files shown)


admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/service/FormBuilderServiceImpl.java
```diff
@@ -1196,7 +1196,7 @@ public EntityForm buildMapForm(MapMetadata mapMd, final MapStructure mapStructur
         } else {
             keyField = new Field().withName("key")
                                 .withFieldType(SupportedFieldType.STRING.toString())
-                                .withFriendlyName("Key");
+                                .withFriendlyName(mapStructure.getKeyPropertyFriendlyName()");
```
charchar: @khanmark Looks like there's a stray `"` (double quotes) here.  Otherwise review/testing is good.

charchar:
👍 Tested and Reviewed

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/service/FormBuilderServiceImpl.java
khanmark: Fixed

PR #1591, opened 2017-07-19
Fix an issue in the admin where the save button can be clicked multiple times leading to duplicated records

phillipuniverse:
Looks like a change was made to the structure, and this method instance was missed in the update. Changed to match Delete and now the spinner/disabling is occurring correctly.


phillipuniverse:
This should likely be rebased onto 5.1 so we can merge it in that line instead of only in 5.2.

mckloc:

jfleschler:
Moved to a new PR #1622

PR #1609, opened 2017-07-21
Fix selectize queries to check data-hydrate if no results

mckloc:
As is, Selectize returns only a page of results. Anything past this cannot be found. So we now add an option from the data-hydrate field of the selectize if the query is empty.


phillipuniverse:
Already fixed by #1613.

PR #1611, opened 2017-07-21
Resize assets on asset grid on each image load

mckloc:
Add a listener to resize the assetGrid whenever an asset image loads, so we match everything to the maximumHeight of all images. This is so, on the last image load, we scale everything appropriately.


mckloc:
This is the wrong collection of commits.

mckloc:
Recreated PR with correct commits at https://github.com/BroadleafCommerce/BroadleafCommerce/pull/1615

PR #1615, opened 2017-07-24
Add a listener to resize images and asset grid on load for each image…

mckloc:
This, combined with [Enterprise PR 945](https://github.com/BroadleafCommerce/Enterprise/pull/945), should fix the assetGrid's sizing both on load and on window resizing.


mckloc:

PR #1616, opened 2017-07-24
Initialize listgrid titles

mckloc:
Listgrid titles were blank because minimum width was not being considered when setting widths. Now, compare with minimum width.


mckloc:

mckloc:
Already merged in.

PR #1618, opened 2017-07-24
Schema.org Metadata

knoxcorner:
This adds [Schema.org](http://schema.org/) metadata through [JSON-LD](https://json-ld.org/). The metadata enhances search engine optimization by providing a way for search engines to recognize the data on a page and display it directly in search results, such as Google's rich card feature.

I've added in metadata for the specific pages:
- Homepage
- Product page
- Category page

All other pages have the same basic metadata.

**Note:** The social media list is not yet implemented.

BroadleafCommerce/QA#2902


knoxcorner:
# Schema.org Metadata

Schema.org provides a standard for structured data. Putting structured data on a webpage allows for external programs, usually crawlers, to see and understand the information the site is displaying. For example, a site showing a recipe would make sense to us- but a crawler has no way to interpret the information on the page. Structured data provides a standard way to mark data in predetermined formats such that a crawler can pick up content on the page.

Structured data can be given in several formats. One form is microdata, which involves adding attributes to existing HTML tags to identify the contents of the tag. RDFa is used in a form similar to microdata, marking up the tags of the page directly. JSON-LD, standing for JSON linked data, stands out from the others because instead of marking up the content on a page, it keeps the information completely isolated from what is on the page. The data is presented in JSON with specific types reflecting those found on Schema.org. 
  
### Why JSON-LD?

JSON-LD is preferred because it does not require marking up the HTML. With the amount of data that can be marked up with Schema.org, HTML tags on a page that were previously simple can become multiple lines long, and the whole page becomes convoluted. Because JSON-LD is isolated from the HTML, it doesn't clog up the source and can be neatly placed at the top or bottom of a page. In addition, with websites that have dynamic webpages, such as the sites running Broadleaf, dynamically adding metadata to tags becomes a hassle. With JSON-LD, it can be much simpler to have the middle tier generate the metadata since it has easier access and then embed that within the page. Google now encourages the use of JSON-LD over the other formats for the same reasons.
 
## Usage

To have the metadata embedded in the page, simply add the tag `<blc:linkedData />` wherever you want the metadata to appear (Google's crawler will find it anywhere in the page).

### Site Navigation Menu

There is additional metadata for site navigation. The script is added as a model attribute `menuLinkedData` anywhere after a `<blc:menu>` tag.

### Configuration

In the *common-shared.properties* file, there are three properties that need to be configured:

 - site.name: The name of the site that will appear in the metadata
 - site.logo: A URL to the logo of the site. Square images work best
 - site.search: A relative path to the search query URL with `{query}` representing the search. I.E. `search?q={query}`.
 
## Implementation Detail

The `LinkedDataProcessor` is what ties the linked data to the presentation tier. Linked data tags will be replaced with the JSON-LD script. The processor first determines what sort of page the tag is on. It handles three specific cases- product, category, and home. All other types of pages contain the default metadata. The processor distinguishes these pages because they each require metadata specific to that page, so some additional handling is required.

Once the `LinkedDataProcessor` determines what type of page is visited, it hands control the the respective service for that type of page. All of the services implement `LinkedDataService` and use `LinkedDataUtil` which provides some methods for getting metadata common to all the pages, including the Organisation, WebSite, Breadcrumbs, and social media sites. Each of the services build up the JSON for their type of page. 

- The `ProductLinkedDataService` fills in information about the product, such as name, description, availability, and pricing for each of the SKUs. It also adds review and rating information if available. It also has a specialized breadcrumb list to demonstrate the categories and subcategories.

- The `CategoryLinkedDataService` provides references to the products in the category.

- The `HomepageLinkedDataService` adds metadata to represent the search feature of the site. This allows for search engines to provide a search bar for the site inside the search results.

- The `DefaultLinkedDataService` provides only the organization, website, and breadcrumb list.

### Site Navigation Menu

The menu metadata is implemented in a service similar to those above. However, it is added as an attribute to the model and does not need to be resolved to the correct page type since it is attached to the menu. It's with the Menu module because the information needed is not in the BroadleafCommerce framework, but inside the menu module. In addition, it contains only metadata regarding the site navigation elements and nothing else.

## Modification and Extension

### Modification

To modify the behavior of the existing services, they can be extended in a new class and overridden. In addition, the `LinkedDataServiceFactory` will need to be overridden and provide the new service implementation.

### Extension

Extension is the same as modification, but the methods `LinkedDataProcessor#resolveDestination` and `LinkedDataProcessor#getDataForDestination` will need to be overridden to resolve the new page type. Additionally, the `LinkedDataServiceFactory` will need a new method to generate your class.

knoxcorner:
[(1) Schema.org Compliance [1]](https://trello.com/c/lE8tGcv2/41-1-schemaorg-compliance-1)

PR #1627, opened 2017-07-31
Refactor checkout flow to enhance the payment stage & better incorporate saved payments

ckittrell:
BroadleafCommerce/QA#2934
BroadleafCommerce/QA#2935


core/broadleaf-framework-web/src/main/java/org/broadleafcommerce/core/web/checkout/model/BillingInfoForm.java
```diff
@@ -104,12 +103,4 @@ public String getPaymentName() {
...
         this.paymentName = paymentName;
     }
-
-    public String getEmailAddress() {
-        return emailAddress;
-    }
-
-    public void setEmailAddress(String emailAddress) {
-        this.emailAddress = emailAddress;
-    }
```
ckittrell: Note: I had added this property in a previously merged PR. This change is simply reverting back to the original state of the BillingInfoForm.

PR #1630, opened 2017-08-02
Add null checks in Category and Product handler mappings

jfleschler:
This pull requests addresses an issue when defining `ResourceHandlerRegistry` handlers.  For example, the following will cause an NPE when you try to hit `/manifest.json`, because the context that is passed in contains nulls.

The `ResourceHandlerRegistry` checks for matches after all other handler mapping have run, so it will need to make it through both the `CategoryHandlerMapping` and the `ProductHandlerMapping` without errors to resolve correctly.

```
    @Override
    public void addResourceHandlers(ResourceHandlerRegistry registry) {
        registry.addResourceHandler("/favicon.ico")
                .addResourceLocations("WEB-INF/favicon.ico");
        registry.addResourceHandler("/manifest.json")
                .addResourceLocations("classpath:/manifest.json");
    }
```

### DIFF
--- a/core/broadleaf-framework-web/src/main/java/org/broadleafcommerce/core/web/catalog/CategoryHandlerMapping.java
+++ b/core/broadleaf-framework-web/src/main/java/org/broadleafcommerce/core/web/catalog/CategoryHandlerMapping.java
@@ -96,19 +96,24 @@ public class CategoryHandlerMapping extends BLCAbstractHandlerMapping {
     }
 
     protected Category findCategoryUsingIdParam(BroadleafRequestContext context) throws ServletRequestBindingException {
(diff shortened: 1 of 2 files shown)


phillipuniverse:
Just wanted to make a note of why this occurs in the first place. The BroadleafRequestFilter doesn't happen on `*.js` or `*.json` requests (URLs with these are explicitly ignored) but Spring MVC still travels through all of the handler mappings. The default ordering for the `ResourceHandlerMapping` (which get created out of the `ResourceHandlerRegistry` above) is `MAX_INT - 1` so it gets resolved last. Product and Category have their ordering at 1 and 2, so they run prior to the `ResourceHandlerMapping`s. Since the BRC gets skipped, you get NPEs.

PR #1632, opened 2017-08-03
SEO Auto Redirects

knoxcorner:
## SEO Auto Redirects
This adds some additional methods for persisting URLHandlers, adds a uniqueness validator to the category page URL in the admin, and adds a way to find URLHandlers from the database by their destination (`newURL`). The actual logic for automatically creating and managing the redirects is in the Enterprise module.

### Additional persistence features
The additional persistence features are necessary because when URLHandlers are loaded from the database and cached, they are recreated as `URLHandlerDTO`s. The DTOs are not entities and cannot be persisted in the normal way. The additional persistence features add a way to save URLHandler's URLs regardless of whether they are entities. 

### Uniqueness validator on category page
The uniqueness validator prevents multiple category page from being mapped to the same URL.

### Search URLHandlers by destination
In order to identify cycles of redirects starting with only the last redirect, there needs to be a way to follow the redirects backwards. The original URLHandler DAO and service did not have a way to do this without loading all URLHandlers which would be too expensive. `findURLHandlersByDestination` allows querying the database in a less expensive manner.


knoxcorner:
[SEO - Auto Redirect](https://trello.com/c/8S6Sw7bY/74-seo-auto-redirect)

admin/broadleaf-contentmanagement-module/src/main/resources/config/bc/jpa/domain/URLHandler.orm.xml
```diff
@@ -39,5 +39,25 @@
         </query>
     </named-query>
 
+    <named-query name="BC_READ_BY_DESTINATION_URL">
+        <query>SELECT handler
+            FROM org.broadleafcommerce.cms.url.domain.URLHandler handler
+            WHERE
+            handler.newURL = :newURL
+        </query>
+    </named-query>
```
ckittrell: We generally want to use `TypedQueries` like the following (https://cl.ly/351q262j3d14) instead of these orm queries

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/validation/UriPropertyValidator.java
```diff
@@ -51,16 +52,19 @@
     protected String ERROR_KEY_CANNOT_END_WITH_SLASH = "uriPropertyValidatorCannotEndWithSlashError";
     protected String ERROR_KEY_CANNOT_CONTAIN_SPACES = "uriPropertyValidatorCannotContainSpacesError";
 
+    @Resource(name = "blSystemPropertiesService")
+    protected SystemPropertiesService systemPropertiesService;
```
ckittrell: We should use an autowired `Environment` instead of this service to gather SystemProperties.

admin/broadleaf-contentmanagement-module/src/main/java/org/broadleafcommerce/cms/url/domain/URLHandlerImpl.java
```diff
@@ -77,14 +69,28 @@
...
+                            validationImplementation = "blUniqueValueValidator",
+                            configurationItems = {
+                                    @ConfigurationItem(
+                                            itemName = ConfigurationItem.ERROR_MESSAGE,
+                                            itemValue = "This URL is already in use. Please provide a unique URL."
+                                    )
+                            }
+                    )
+            }
+    )
```
ckittrell: Move the `@AdminPresentation` annotation below the column & index annotations

PR #1635, opened 2017-08-04
Modified OrderPaymentService to utilize OrderService's remove payment method

ncrum:
This goes along with the attached issue. Essentially removed the new method to remove payments and used the one that exists on OrderService


ncrum:
[Two payments when checking out while creating a saved payment](https://trello.com/c/SZdzZLDq/143-two-payments-when-checking-out-while-creating-a-saved-payment)

PR #1638, opened 2017-08-07
Upgraded Log4J dependency and references to Log4J2 and made the dependency optional

nathandmoore:
>Spring Boot's default logger is Logback and we should be able to work with that rather than requiring Log4J.

## Changes
- Upgraded dependency on Log4J to Log4J2.
    - Made the Log4J dependency optional
- Deprecated `RuntimeLog4jCofigurer` since Log4J is not the default logger and this class was intended for internal use.
- Update javadoc for `ProcessDetailLogger.javaProcessDetailLogger` to use Logback configuration as the example of a rolling daily log instead of Log4J
- Switched ESAPI to default to JUL instead of Log4J 
    - We don't use ESAPI logging, so should be a superficial change
- Excluded Log4J from the ESAPI dependency to prevent HSQLDB from automatically configuring itself to use Log4J
    - HSQLDB will use Log4J if it finds it on the classpath.


nathandmoore:
[(0.5) Investigate and potentially remove any Log4j/SLF4j classes](https://trello.com/c/Pv6JQPtZ/56-05-investigate-and-potentially-remove-any-log4j-slf4j-classes)

common/src/main/resources/esapi/ESAPI.properties
```diff
@@ -92,8 +92,8 @@ ESAPI.Executor=org.owasp.esapi.reference.DefaultExecutor
 ESAPI.HTTPUtilities=org.owasp.esapi.reference.DefaultHTTPUtilities
 ESAPI.IntrusionDetector=org.owasp.esapi.reference.DefaultIntrusionDetector
 # Log4JFactory Requires log4j.xml or log4j.properties in classpath - http://www.laliluna.de/log4j-tutorial.html
-ESAPI.Logger=org.owasp.esapi.reference.Log4JLogFactory
-#ESAPI.Logger=org.owasp.esapi.reference.JavaLogFactory
+#ESAPI.Logger=org.owasp.esapi.reference.Log4JLogFactory
```
phillipuniverse: You should never have commented out code committed since we can go back in source control and see what used to be there. This line should be removed not commented out.

nathandmoore: Just following the pattern set by ESAPI in their own default properties.

nathandmoore:
The dependencies shouldn't conflict at all since the main packages are different: `org.apache.log4j -> org.apache.logging.log4j`. I'll add the old fashioned ones back.

PR #1642, opened 2017-08-07
Registration Password only accepts 6 or more non-space characters

mckloc:
Now matches the change password form's requirement


mckloc:
[Create account password requirements don't match update password - validation (PrivateDemo also)](https://trello.com/c/EAlmo6mJ/148-create-account-password-requirements-dont-match-update-password-validation-privatedemo-also)

PR #1645, opened 2017-08-08
Fixed issue where items were unable to be added to wishlist when out of stock

ncrum:


ncrum:
[Cannot add an item to a wishlist if it is out of stock](https://trello.com/c/yRl5CNbh/122-cannot-add-an-item-to-a-wishlist-if-it-is-out-of-stock)

PR #1646, opened 2017-08-09
Order History - Searching, Pagination, and Date Filtering

knoxcorner:
## Enhance Order History
This enhances order history by adding a basic search feature, date filtering, and pagination support for the `BroadleafOrderHistoryController`. The majority of the logic for filtering is in `OrderHistoryService` since the majority of the features are shared between the `BroadleafOrderHistoryController` and the `BroadleafAccountOrderHistoryController` located in the Account module.

### Pagination
The enhancement allows for pagination of orders on the order history page. Previously, all orders for would be listed on a single page regardless of the number of orders. Now, the orders are broken into pages.

### Date filtering
The orders can now be filtered within a date range. For example, by changing the start date to 6 months ago and leaving the end date at the current date, you can retrieve all orders within the past 6 months.

### Searching
Orders can be searched using the order number, a product SKU, or a partial product name.

**Note:** The search implementation is very basic and will be replaced in the near future with Solr searching. Date filtering will also likely be done through Solr.

### DIFF
--- a/core/broadleaf-framework-web/src/main/java/org/broadleafcommerce/core/web/controller/account/BroadleafOrderHistoryController.java
+++ b/core/broadleaf-framework-web/src/main/java/org/broadleafcommerce/core/web/controller/account/BroadleafOrderHistoryController.java
@@ -17,25 +17,17 @@
  */
 package org.broadleafcommerce.core.web.controller.account;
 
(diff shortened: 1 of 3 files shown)


knoxcorner:
[Order History Material UI update](https://trello.com/c/XhaDH62b/139-order-history-material-ui-update)

PR #1647, opened 2017-08-09
Add State/Province/Region, PostalCode, and PhoneNumber Empty/Whitespace check to Common Address Validator.

mckloc:
Requires these three fields to have something in them when submitting Addresses that use this default validator.


mckloc:
https://trello.com/c/oL2EJwMk/152-my-account-issues

Fixing issue from above card:

PR #1648, opened 2017-08-10
Added ability to have an amount off offer split among the matching order items

cja769:
This is similar to an amount off order item discount except that that the value of the offer is split among all matching offers. So a $5 split amount offer would always adjust the order subtotal by $5 regardless of how many items are matched. This is a useful offer when considering returns since a $5 off order promo would effectively be split among all order items in the order when considering returns, whereas a split amount order item offer will already have the discount split among the targeted order items

Additionally this was a feature requested by O'Reilly.

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/offer/service/OfferServiceUtilitiesImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/offer/service/OfferServiceUtilitiesImpl.java
@@ -294,8 +294,18 @@ public class OfferServiceUtilitiesImpl implements OfferServiceUtilities {
         }
         itemOffer.setItemsApplied(itemsApplied);
         if (CollectionUtils.isNotEmpty(confirmedDetails)) {
+            Money amountApplied = new Money(BigDecimal.ZERO);
             for (PromotableOrderItemPriceDetail detail : confirmedDetails) {
                 applyOrderItemAdjustment(itemOffer, detail);
+                for (PromotableOrderItemPriceDetailAdjustment adjustment : detail.getCandidateItemAdjustments()) {
+                    if (adjustment.getOfferId().equals(itemOffer.getOffer().getId())) {
+                        // Since it's an amount off it doesn't matter which one is used. The sale and retail should be the same
+                        amountApplied = amountApplied.add(adjustment.getRetailAdjustmentValue().multiply(detail.getQuantity()));
+                    }
+                }
+            }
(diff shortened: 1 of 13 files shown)


bpolster:
@cja769 This will take some time to review.    I'd like to have a conversation with you about the source of the requirement (e.g. customer / internal).    It overlaps with something that we've put thought into before regarding discount allocation.

cja769:
As a note I think the logic around actually splitting the offer between the order items works pretty well and I'm pretty sure I did it correctly. What I'm unsure of is the process I made to fix the rounding errors. Specifically when I had to create a new price detail since quantity was more than 1 so in order to only adjust the subtotal by 1 penny.

For example I have a split offer that applied to one line item with quantity 3. The offer is for $5. This, by defaults, creates 1 adjustment that targets all 3 quantity of the line item. The adjustment value would be $1.6666 rounded to $1.67. $1.67 * 3 = $5.01 so I need to adjust the price by a penny. If I simply change the adjustment to $1.66 then the discount would come out to $1.66 * 3 = $4.98. In order to fix this I created a new adjustment that's a copy of the original adjustment but with quantity 1 and value $1.67. I then change the original adjustments quantity to 2 so that it doesn't apply to all of the items.

The problem with my above example is it creates 2 adjustments with the same offer. By default the offer engine tries to merge like adjustments into 1 adjustment which would then result in 1 adjustment targeting all 3 items with either a value of $1.67 or $1.66 (which are both wrong). To fix this issue I modified `PromotableOrderItemPriceDetailImpl` to have a `isOverride` property and I use that property, if it's set to true, to generate the key in `PromotableOrderItemPriceDetailImpl#buildDetailKey`. Unfortunately since I can't add domain I can't add a `isOverride` field to `OrderItemPriceDetail`. 

The above makes `OrderOfferProcessor#synchronizeItemPriceDetails` act a little odd since none of the current `OrderItemPriceDetail`s find a matching `PromotableOrderItemPriceDetail`. I think, in general, this is fine since it just reuses unmatched `OrderItemPriceDetail`s to create the `PromotableOrderItemPriceDetail`s that it deemed didn't already exist.

jefffischer:
I haven't reviewed this or anything, but I want to clarify a statement you made. You can in fact add domain as long as it's optionally weaved based on a property that's false by default. Presumably, clients that need or want the feature will be willing to set the property to true and add the empty column to the table. Other clients will not be impacted.

________________________________

cja769:
Ah right I forgot about that option. Alternatively this change can simply target 5.3.x and I can make these changes in O'Reilly's code directly.

bpolster:
We are going to table this for now and @cja769 is going to make the changes locally for O'Reilly.

PR #1656, opened 2017-08-15
Fixed issue where rating service was incorrectly counting duplicates for the average rating

ncrum:
This issue really only comes into play when trying to leave a new review to replace an existing one, which is not in our normal demo site, but is possible using the API.


ncrum:
[React Material Kit Updates](https://trello.com/c/f7TLEggV/202-react-material-kit-updates)

PR #1640, opened 2017-08-07
[Issue#1639]

linuxidefix:
Displayed value of the property's name must be unescaped.
#1639

Issue #1639, opened 2017-08-07
Accented characters are not correctly displayed in admin


linuxidefix:
Version : 5.2.X
- Tab names et Group names are ok.
- Property name with html characters are not displayed correctly.

![blcadminaccentedcharacters](https://user-images.githubusercontent.com/836182/29033545-b07f82b2-7b95-11e7-8bd1-2a3c1cdf9d13.png)

- problem comes from template field template.

linuxidefix:
Fix in pull request #1640
![blcadminaccentedcharactersfixed](https://user-images.githubusercontent.com/836182/29034022-2cb3f22c-7b97-11e7-9972-1f4b05691aa6.png)

PR #1658, opened 2017-08-15
Allow UTF-8 characters (like accent marks) in message bundles

phillipuniverse:
This fixes issues in our message bundles with translations that was referenced in BroadleafCommerce/DemoSite#10 and also refines #1639 and #1640 by allowing translations to use their actual accented characters and not html-entities equivalencies.


phillipuniverse:
[(.25) DemoSite#10 - cannot put characters with accent marks in message bundles](https://trello.com/c/JAm2dobX/120-25-demosite10-cannot-put-characters-with-accent-marks-in-message-bundles)

phillipuniverse:
I have not yet verified this, should be easy enough; just add an accent mark character (like ñ) to an admin message and a site message in one of the properties file. When I pasted it in Eclipse it used the UTF-8 encoded character (like `\u4341`) but I'm not sure if that works.

cja769:
Confirmed
https://cl.ly/2h0O263L3G0F
https://cl.ly/2O200X1b2S0V

Also ñ is \u00f1 in eclipse

PR #1663, opened 2017-08-21
Add method to the framework to query products that use a specific product option

Jerry77oz:
BroadleafCommerce/QA#2972 

- Add support to query products that use a specific product option with a start `position` and `pageSize`
- Added query to do a count of all the products using a product option


danielcolgrove:
@Jerry77oz Is there any possibility that the query for Products using a particular ProductOption could return a very large result set?  Just thinking of potential performance implications with this query,

PR #1669, opened 2017-08-27
Fill In Saved Customer Addresses

jmitash:
This changes the approach for loading saved customer addresses. Previously, the addresses were loaded from the manage addresses page with a `GET` request. With dynamic regional forms, one Ajax request would be needed to load the form for the right country, and another to load the address information. Rather than use nested Ajax requests, I've added a `customerAddressId` to the related forms and use it to fill in the saved address on the server's side.


knoxcorner:
[(0.5) i18n DynamicRegionalForms](https://trello.com/c/wqJxir7b/58-05-i18n-dynamicregionalforms)

PR #1673, opened 2017-08-30
Sku#getMultiValueSkuAttributes has inaccurate return type

oleksiimiroshnyk:
BroadleafCommerce/QA#2980


nathandmoore:
@oleksiimiroshnyk Do we know how Buckle is using this method? Also, the return type needs to be `Map<String, **Collection**<SkuAttribute>>` not `Map<String, **List**<SkuAttribute>>`. I think we can also go ahead and undeprecate `getSkuAttributes()`.

oleksiimiroshnyk:
@nathandmoore I didn't get any response on how Buckle is using this.

oleksiimiroshnyk:
@nathandmoore I updated PR with collection instead of list

PR #1677, opened 2017-09-05
Enabled entityManager.find() catalog validation when grabbing a page entity.

charchar:
Prevents page variables from unintentionally appearing across catalogs.

Addresses BroadleafCommerce/QA#3021.

### DIFF
--- a/admin/broadleaf-contentmanagement-module/src/main/java/org/broadleafcommerce/cms/web/PageHandlerMapping.java
+++ b/admin/broadleaf-contentmanagement-module/src/main/java/org/broadleafcommerce/cms/web/PageHandlerMapping.java
@@ -82,7 +82,7 @@ public class PageHandlerMapping extends BLCAbstractHandlerMapping {
 
             } finally {
                 if (internalValidateFindPreviouslySet) {
-                    BroadleafRequestContext.getBroadleafRequestContext().setInternalValidateFind(true);
+                    BroadleafRequestContext.getBroadleafRequestContext().setInternalValidateFind(false);
                 }
             }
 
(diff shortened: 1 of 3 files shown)


admin/broadleaf-contentmanagement-module/src/main/java/org/broadleafcommerce/cms/web/PageHandlerMapping.java
```diff
@@ -69,7 +69,22 @@ protected Object getHandlerInternal(HttpServletRequest request) throws Exception
...
+            try {
+                if (!BroadleafRequestContext.getBroadleafRequestContext().getInternalValidateFind()) {
+                    BroadleafRequestContext.getBroadleafRequestContext().setInternalValidateFind(true);
+                    internalValidateFindPreviouslySet = true;
+                }
+                page = pageService.findPageByURI(context.getLocale(), requestUri, buildMvelParameters(request), context.isSecure());
+
+            } finally {
+                if (internalValidateFindPreviouslySet) {
+                    BroadleafRequestContext.getBroadleafRequestContext().setInternalValidateFind(true);
```
nathandmoore: methinks this should be false not true

admin/broadleaf-contentmanagement-module/src/main/java/org/broadleafcommerce/cms/web/controller/BroadleafPageController.java
```diff
@@ -77,7 +77,23 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
...
+            if (!BroadleafRequestContext.getBroadleafRequestContext().getInternalValidateFind()) {
+                BroadleafRequestContext.getBroadleafRequestContext().setInternalValidateFind(true);
+                internalValidateFindPreviouslySet = true;
+            }
+
+            extResult = templateOverrideManager.getProxy().getOverrideTemplate(erh, page);
+        } finally {
+
+            if (internalValidateFindPreviouslySet) {
+                BroadleafRequestContext.getBroadleafRequestContext().setInternalValidateFind(true);
```
nathandmoore: again, should it not be false if you previously set it to true in this same scope? Otherwise, it's already true

admin/broadleaf-contentmanagement-module/src/main/java/org/broadleafcommerce/cms/web/PageHandlerMapping.java
charchar: Good catch, fixed along with below

PR #1679, opened 2017-09-06
When creating a new entities in the admin with a generated URL, the URL keeps a period appended at the end

oleksiimiroshnyk:
BroadleafCommerce/QA#3000


admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/entityForm-generatedUrl.js
```diff
@@ -86,8 +86,10 @@
 
             if (options != null && options.allowSlash) {
                 return val.replace(/ /g, BLC.systemProperty.urlFragmentSeparator).replace(/[^\w\s-_\/]/gi, '').toLowerCase() + '.' + valPostFix;
```
phillipuniverse: Shouldn't the `if(ValPostFix)` check be up here to? Or at the end regardless? The if statement on this line also appends it onto the end.

oleksiimiroshnyk: probably true
Slightly changed aproach of checking this...

PR #1703, opened 2017-09-18
Added ability to force customers to change their passwords

nathandmoore:
## Feature Overview

During a site conversion or security breach or a matter of routine security policy, it may be necessary to require users to change their passwords. This feature allows for implementors to force customers to reset their passwords when they try to log in if `Customer#passwordChangeRequired` is `true`. The flow is the same as the forgot password flow except in being mandatory and using different verbiage and view names.

## Changes

- Added javadoc explanation for the usage of Customer#passwordChangeRequired
- Extended the capabilities of the BroadleafAuthenticationFailureHandler so that you can add view mappings for various exceptions that can occur during the auth flow.
- Add ability to force a customer to change their password OOB. Works essentially the same way as forgot password, but with different views and verbiage.  
    - Generified the request attribute name for username, when attached for password, reset after forgetting password or forcing being forced to reset password 
    - Added method to send the forced-password-change notifications. Simply does what the same as for sending the forgot-password notification 
    - Set the customer#passwordChangeRequired back to false once they've changed it

## Testing (in ReferenceSite)

1. Take an existing user and modify the `passwordChangeRequired` property in the DB to true.
2. Attempt to login.
3. The user should be presented with the new password-change-required screen with appropriate copy.
4. The user requests to change their password (ideally email is prepopulated from the prior screen but could be edited).
5. User receives an email with the token (similar to existing forgot password behavior)
6. User changes their password and is logged in on success
7. User logs out and logs back in with the new password.


nathandmoore:
[(1) OOB ability for forced password reset](https://trello.com/c/HK2h50B4/94-1-oob-ability-for-forced-password-reset)

PR #1707, opened 2017-09-21
Enhanced FieldDefinition fields with `@AdminPresentation` so that they could be overridden, and added more field types to DynamicSupportedFieldTypes enumeration so that more basic fields are available in the admin.

nathandmoore:
## Changes
- DynamicSupportedFieldTypes is used by PageType in the admin to determine what FieldDefinitions can be used by Pages to create PageFields. However, it was missing several commonly used SupportedFieldTypes that have now been added:
    - Boolean
    - Integer
    - Decimal
- Added `@AdminPresentation(excluded = true)` for all FieldDefinition fields that previously had no AdminPresentation so that they can be overridden with MetaDataOverrides.
- Made `FieldDefinition#fieldType` accessible as a `DynamicSupportedFieldType` so as to all access to friendlyType or other options.


nathandmoore:
[(8) Generic form type pages [2.5]](https://trello.com/c/6xVkjkTD/96-8-generic-form-type-pages-25)

PR #1722, opened 2017-10-16
Added "Within x days" operator for date rules that checks if the field date is within the specified number of days from the date at time of evaluation.

charchar:
Addresses BroadleafCommerce/QA#3071.


common/src/main/java/org/broadleafcommerce/common/rule/MvelHelper.java
```diff
@@ -80,7 +84,11 @@ public static Object convertField(String type, String fieldValue) {
             if (type.equals(SupportedFieldType.BOOLEAN.toString())) {
                 return Boolean.parseBoolean(fieldValue);
             } else if (type.equals(SupportedFieldType.DATE.toString())) {
-                return FormatUtil.getTimeZoneFormat().parse(fieldValue);
+                try {
```
bpolster: Change to use a length check instead of a  ... getTimeZoneFormat().length() instead of relying on an exception.    Need to try and remove and recreate and then add the length check.

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/rulebuilder/DataDTOToMVELTranslator.java
```diff
@@ -324,6 +324,24 @@ protected void buildExpression(ExpressionDTO expressionDTO, StringBuffer sb, Str
                 }
                 break;
             }
+            case WITHIN_DAYS: {
+
+                try {
+                    Integer.parseInt(value[0] + "");
+                } catch (Exception e) {
```
bpolster: Try to do this without a catch Exception block.

common/src/main/java/org/broadleafcommerce/common/rule/MvelHelper.java
```diff
@@ -98,7 +106,20 @@ public static Object toUpperCase(String value) {
         }
         return value.toUpperCase();
     }
-    
+
+    public static String currentTime() {
+        Calendar calendar = Calendar.getInstance();
```
bpolster: Use SystemTime to get the calendar

common/src/main/java/org/broadleafcommerce/common/rule/MvelHelper.java
```diff
@@ -98,7 +106,20 @@ public static Object toUpperCase(String value) {
...
+
+        return FormatUtil.getTimeZoneFormat().format(calendar.getTime());
+    }
+
+    public static String subtractFromCurrentTime(int days) {
+        Calendar calendar = Calendar.getInstance();
```
bpolster: Use system time to get the calendar.

PR #1723, opened 2017-10-16
Added in better support for Email and SMS notifications

ncrum:
Migration is required for client's looking to upgrade, but use the same older email infos. These requirements are detailed in the Migration Notes.

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/rule/MvelHelper.java
+++ b/common/src/main/java/org/broadleafcommerce/common/rule/MvelHelper.java
@@ -256,7 +256,7 @@ public class MvelHelper {
      */
     public static Map<String, Object> buildMvelParameters() {
         Map<String, Object> mvelParameters = new HashMap<String, Object>();
-       BroadleafRequestContext brc = BroadleafRequestContext.getBroadleafRequestContext();
+        BroadleafRequestContext brc = BroadleafRequestContext.getBroadleafRequestContext();
         if (brc != null && brc.getRequest() != null) {
            TimeDTO timeDto = new TimeDTO(SystemTime.asCalendar());
             HttpServletRequest request = brc.getRequest();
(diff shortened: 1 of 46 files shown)


common/src/main/java/org/broadleafcommerce/common/rule/MvelHelper.java
```diff
@@ -269,8 +269,15 @@ public static void setTestMode(boolean testMode) {
                 for (String mapKey : blcRuleMap.keySet()) {
                     mvelParameters.put(mapKey, blcRuleMap.get(mapKey));
                 }
-           }
-       }
+            }
```
elbertbautista: Should be removed

common/src/main/java/org/broadleafcommerce/common/web/BroadleafRequestContext.java
```diff
@@ -127,6 +127,7 @@ public static BroadleafCurrency getCurrency() {
     protected List<Long> explicitCatalogs = new ArrayList<>();
     protected Site currentProfile;
     protected Boolean ignoreSite = false;
+    protected Map<String, Object> ruleMap = new HashMap<>();
```
elbertbautista: extraneous - can be removed

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/checkout/service/workflow/CompleteOrderActivity.java
```diff
@@ -57,9 +50,6 @@ public CompleteOrderActivity(@Qualifier("blCompleteOrderRollbackHandler") Comple
         seed.getOrder().setOrderNumber(determineOrderNumber(seed.getOrder()));
         seed.getOrder().setSubmitDate(determineSubmitDate(seed.getOrder()));
 
-        OrderSubmittedEvent event = new OrderSubmittedEvent(seed.getOrder().getId(), seed.getOrder().getOrderNumber());
```
elbertbautista: Just make sure that this still works with indexing Orders and Fulfillment Orders in SOLR

PR #1746, opened 2017-11-15
Image corruption can occur when multiple treatments for the same image are requested concurrently

hasheniuk:
BroadleafCommerce/QA#3094
concurrent file writing component


hasheniuk:
@jefffischer 
I want to clear the map in some way. When the file already moved, it doesn't need the lock object anymore. I used WeakHashMap to ensure it would be cleared by GC. But WeakHashMap hasn't concurrent implementation.

jefffischer:
I guess I don't have a huge preference either way, as this section of code should only be hit for new or previously unrequested graphics.

PR #1773, opened 2018-01-19
Fixed issue where BLC was swapping Solr cores when there was an error during reindexing.

ktisdell:
https://github.com/BroadleafCommerce/QA/issues/3171 QA-3171 issue where BLC was swapping Solr cores when there was an

error during reindexing, which leaves the live site with an empty index.


phillipuniverse:
@ktisdell actually, there are some operations that set up some things on threadlocal that actually need to clean up in a `finally{}` block (e.g. MultiTenant and sandboxing operations). Agree that swapping the Solr cores should definitely not happen in a `finally{}` but the block should still exist, and maybe we have a new method that does a cleanup within the `try{}`

PR #1787, opened 2018-02-12
Added ability to handle extended map entity xrefs through the admin

jfleschler:
This PR handles the case when an Xref entity is extended and participates in an `@AdminPresentationMap`.

The changes allow for adding properties from extensions of the base class.

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/MapStructurePersistenceModule.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/MapStructurePersistenceModule.java
@@ -143,7 +143,7 @@ public class MapStructurePersistenceModule extends BasicPersistenceModule {
                     );
                 } else {
                     String valueClassName = mapStructure.getValueClassName();
-                    Class<?>[] entities = persistenceManager.getDynamicEntityDao().getAllPolymorphicEntitiesFromCeiling(Class.forName(valueClassName));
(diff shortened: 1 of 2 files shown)


admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/MapStructurePersistenceModule.java
```diff
@@ -142,9 +142,11 @@ public void updateMergedProperties(PersistencePackage persistencePackage, Map<Me
                         MergedPropertyType.MAPSTRUCTUREVALUE
                     );
                 } else {
+                    String valueClassName = mapStructure.getValueClassName();
+                    Class<?>[] entities = persistenceManager.getDynamicEntityDao().getAllPolymorphicEntitiesFromCeiling(Class.forName(valueClassName));
```
jefffischer: Curious why this entity array retrieval is written differently than the others

jefffischer (changes requested):

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/MapStructurePersistenceModule.java
jfleschler: It was the first place I changed and forgot to go back and fix it.  I've made the change to match the other locations.

jefffischer:
Seems fine as long as the code continues to work with existing maps.

jefffischer (approved):

PR #1792, opened 2018-02-16
Added Order and OrderItem Methods to support Export module

michaelkloc:
readNumber and readBatch methods added to Order and OrderItem Daos

[https://github.com/BroadleafCommerce/Export/pull/11](https://github.com/BroadleafCommerce/Export/pull/11)

[https://github.com/BroadleafCommerce/QA/issues/3150](https://github.com/BroadleafCommerce/QA/issues/3150)

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/dao/OrderDaoImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/dao/OrderDaoImpl.java
@@ -462,24 +462,7 @@ public class OrderDaoImpl implements OrderDao {
         List<Order> orders = query.getResultList();
         return orders != null ? orders : new ArrayList<Order>();
     }
-    
-    @Override
-    public List<Order> readBatchOrders(int start, int pageSize) {
-        CriteriaBuilder builder = em.getCriteriaBuilder();
-        CriteriaQuery<Order> criteria = builder.createQuery(Order.class);
-        Root<OrderImpl> order = criteria.from(OrderImpl.class);
-        criteria.select(order);
-        criteria.where(builder.equal(order.get("status"),OrderStatus.SUBMITTED.getType()));
-        //criteria.add(Restrictions.eq("OrderStatus", "SUBMITTED"));
-        TypedQuery<Order> query = em.createQuery(criteria);
-        query.setFirstResult(start);
-        query.setMaxResults(pageSize);
-        query.setHint(QueryHints.HINT_CACHEABLE, true);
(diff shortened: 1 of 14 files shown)


core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/dao/OrderDaoImpl.java
```diff
@@ -460,4 +462,30 @@ protected Long getDatabaseOrderLockTimeToLive() {
         List<Order> orders = query.getResultList();
         return orders != null ? orders : new ArrayList<Order>();
     }
+    
+    @Override
+    public List<Order> readBatchOrders(int start, int pageSize) {
```
elbertbautista: This method seems redundant. There's already a readBatchOrders method where you can pass in a list of qualified statuses already. That should be used instead of this one

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/dao/OrderItemDaoImpl.java
```diff
@@ -144,4 +144,27 @@ public OrderItemPriceDetail initializeOrderItemPriceDetails(OrderItem item) {
...
+    @Override
+    public Long readNumberOfOrderItems() {
+        CriteriaBuilder builder = em.getCriteriaBuilder();
+        CriteriaQuery<Long> criteria = builder.createQuery(Long.class);
+        criteria.select(builder.count(criteria.from(OrderItemImpl.class)));
+        TypedQuery<Long> query = em.createQuery(criteria);
+        return query.getSingleResult();    }
+
+    @Override
+    public List<OrderItem> readBatchOrderItems(int start, int count) {
```
elbertbautista: for the purposes of an export - seems like this method also needs to be qualified as well. i.e. if you only export submitted orders, you would likely only want to export order items that belong to those submitted orders as well (not all order items). This should consider order statuses like the order method

elbertbautista (changes requested):

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/dao/OrderDaoImpl.java
michaelkloc: Removed

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/dao/OrderItemDaoImpl.java
michaelkloc: Added qualification

elbertbautista (approved):

PR #1795, opened 2018-02-20
Allow loading of CSS resource bundles to be deferred

jefffischer:
BroadleafCommerce/QA#3223

Linked CSS will block initial rendering of a page until the browser fully loads the CSS and parses it. A strategy for faster page loads includes inlining enough CSS to accurately render above-the-fold content (critical CSS) and deferring the load of the remainder. This enhancement allows CSS bundle declarations in Thymeleaf to declare a `defer=true` attribute to formulate the resulting CSS bundle is such a way as to defer its loading and processing until post initial content painting.


danielcolgrove:
This is already handled in https://github.com/BroadleafCommerce/BroadleafCommerce/pull/1900

Issue #1793, opened 2018-02-19
Compilation error on OptimisticLockUtils class in 5.2.x branch 


hamza00111:
Hi, 

When I import the broadleaf framework I find a compilation error of a missing declaration of a variable 'log'.

danielcolgrove:
@hamza00111 We are using Lombok for the logging in that Class.  The annotation @CommonLog handles the inclusion of the `log`.

See https://projectlombok.org/api/lombok/extern/apachecommons/CommonsLog.html

PR #1798, opened 2018-02-22
Add missing LOG variable to correct the compilation error #1793

hamza00111:
When I fork the project I find missing Log variable in OptimisticLockUtils class.

So I created a log variable of apache commons logging as a field.


zaghost:
danielcolgrove explained the problem in #1793

PR #1799, opened 2018-02-22
Added support for setting map values to null through fields such as custom fields

cja769:
Currently you're unable to unset values set through CustomFields (or at least for something like ProductAttributes). This was because the BasicPersistenceModule would not run through custom FieldPersistenceProviders if the value was null and the way the default FieldPersistenceProvider handled null map values would result in no change to the database. The change made was to add a function to FieldPersistenceProviders to ask if they supported populating null values and if so we would use it to handle the null value. If no FieldPersistenceProviders are found meeting that criteria the default FieldPersistenceProvider is used like it was before.

Specifically referring to the MapFieldPersistenceProvider, the entry in the map still remains but it's value is set to null.

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/BasicPersistenceModule.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/BasicPersistenceModule.java
@@ -336,8 +336,6 @@ public class BasicPersistenceModule implements PersistenceModule, RecordHelper, 
                 }
                 String value = property.getValue();
                 if (metadata != null) {
-                    Boolean mutable = metadata.getMutable();
-                    Boolean readOnly = metadata.getReadOnly();
 
                     if (metadata.getFieldType().equals(SupportedFieldType.BOOLEAN)) {
                         if (value == null) {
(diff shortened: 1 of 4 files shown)


cja769:
@jefffischer As mentioned in the description of this pull request, the map values are simply set to null via the MapFieldPersistenceProvider. Do you know of a way to actually remove the entry from the map completely. I was having issues figuring out how to do this because of how `MapFieldPersistenceProvider#updateAssignableValue` works. It calls
```
populateValueRequest.getFieldManager().setFieldValue(assignableValue, manyToField,
                    middleInstance);
```
where the assignableValue is a ProductAttribute. I don't think we can simply set assignableValue to null because 1. I don't think it would delete the entry and 2. I'm fairly certain there would be NullPointerExceptions.

I looked at how `FieldManager#handleMapFieldPopulation` handles the population of a null field but this is the code that currently runs today and does not successfully delete the entry from the map. it has this piece
```
if (newValue == null) {
            Object currentValue = map.get(mapKey);
            if (currentValue != null && currentValue instanceof ValueAssignable) {
                ((ValueAssignable) currentValue).setValue(null);
            } else {
                map.remove(mapKey);
            }
        } else {
            map.put(mapKey, newValue);
        }
```
but for ProductAttributes the map has many duplicate entries and after returning from this function back to the `BasicPersistenceModule` the product instance still has the entry we were trying to remove.

jefffischer:
I think rather than trying to modify fieldmanager to interpret a null as a delete, there should probably by something in mapstructurepersistenceprovider.update to detect and call to #delete instead. Hope that made sense - I’m not at a computer.

cja769:
@jefffischer Oh okay that makes a lot of sense. Thanks!

cja769:
@jefffischer I added the change to delete the ValueAssignable instead of simply updating and it's working as intended. When the CustomField entity field is being set to null the ProductAttribute is now being removed and instead of being updated.

PR #1815, opened 2018-03-02
Using @Resource to resolve the fileExtensionWhitelist dependency in ProcompressedArtifactFilter

danielcolgrove:
Resolves https://github.com/BroadleafCommerce/QA/issues/3237

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/web/util/PrecompressedArtifactFilter.java
+++ b/common/src/main/java/org/broadleafcommerce/common/web/util/PrecompressedArtifactFilter.java
@@ -20,6 +20,7 @@ package org.broadleafcommerce.common.web.util;
 import org.apache.commons.lang.ArrayUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.core.env.Environment;
 import org.springframework.web.filter.GenericFilterBean;
 
@@ -47,9 +48,9 @@ public class PrecompressedArtifactFilter extends GenericFilterBean {
 
     private boolean useWhileInDefaultEnvironment = true;
 
-    @Resource(name = "blPrecompressedArtifactFileExtensionWhitelist")
+    @Value("#{blPrecompressedArtifactFileExtensionWhitelist}")
     List<String> fileExtensionWhitelist;
-
+    
     @Resource
     Environment environment;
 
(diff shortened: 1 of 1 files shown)


Jerry77oz (approved):

phillipuniverse:
@danielcolgrove this does not work and is unsafe to merge in. If you use `@Resource` here in a filter, starting up with Spring Boot will fail since filters are registered directly with the servlet container, and the servlet container (like Tomcat) will try to inject into the `@Resource` by doing a JNDI lookup. See BroadleafCommerce/BroadleafCommerce#1771 and BroadleafCommerce/BroadleafCommercePrivate#958.

phillipuniverse (dismissed):
@danielcolgrove this does not work and is unsafe to merge in. If you use `@Resource` here in a filter, starting up with Spring Boot will fail since filters are registered directly with the servlet container, and the servlet container (like Tomcat) will try to inject into the `@Resource` by doing a JNDI lookup. See BroadleafCommerce/BroadleafCommerce#1771 and BroadleafCommerce/BroadleafCommercePrivate#958.

Jerry77oz:
@phillipuniverse we saw the issue where the change (and reason) why it got changed to `@Autowired` but the issue we were struggling with was, why was it failing when you were testing but it was working for Casey (Icon).

Casey was booting up with `@Resource` (on `5.2.3-M6`) without exceptions and now that he upgraded to `5.2.3-GA` he can't start up because the `@Autowired` is failing to resolve the `List<String>`.

phillipuniverse:
@danielcolgrove did a bit of research here and it seems like this is a common problem with primitive types like this. Random person on the internet from https://tamasgyorfi.net/2014/04/02/spring-injection-with-collections/ suggests to change it to this:

```java
@Value("#{blPrecompressedArtifactFileExtensionWhitelist}")
List<String> fileExtensionWhitelist;
```

The `#{}` expression in SpEL means look up a bean named whatever is inside.

The issue seems to be that when you `@Autowire` a collection, Spring treats that as trying to inject every bean of that type in the generic. So Spring looks for every `String` that it manages and then tries to filter that list down to the specific name in the `@Qualifier`. I guess since we don't have any `String` beans, it fails that it cannot find it. The mismatch is that we have a `List<String>` bean and it actually looks for a `String` bean.

Likely that we need to spot-check the other changes that I made in the referenced issues.

phillipuniverse:
@danielcolgrove @Jerry77oz 1 more thing on this, from the main man Juergen himself: https://jira.spring.io/browse/SPR-12180?focusedCommentId=107840&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-107840

Jerry77oz:
Juergen has spoken!

This was the only case of `List<String>` in the PR, all the others were using specific classes with the `@Autowired`.

danielcolgrove:
@Jerry77oz These changes worked for Casey.  We should be able to merge.

PR #1822, opened 2018-03-08
Offer Validation and Rulebuilder Errors 

jmitash:
- Resize the quantity input in the offer rule builder: The input should now expand to fit the text when multiple digits are entered
- Fix JavaScript errors when changing the quantity input: Previously, changing the quantity was throwing JS errors because the `data-orig-val` attribute was added with an incorrect name. Additionally, the `data-orig-val` would have been incorrect since refocusing the input was changing the value, so refocusing after changing it would set the `data-orig-val` to the current value.
- Wrap server errors in a modal: Previously the errors were being shown as a modal without actually being a modal. This was breaking the page as the black overlay would pop up but no message was visible.
- Added validation to the offer 'value' field to make it required. Since the field is dynamically added, no validators were being added to it for some use-cases. Saving without entering a value was causing an exception at the data layer and also causing the black screen described above.

BroadleafCommerce/QA#3235

### DIFF
--- a/admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/server/service/handler/OfferCustomPersistenceHandler.java
+++ b/admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/server/service/handler/OfferCustomPersistenceHandler.java
@@ -17,6 +17,21 @@
  */
 package org.broadleafcommerce.admin.server.service.handler;
 
+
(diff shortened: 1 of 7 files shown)


jmitash:
[(1.75) Issues creating new promotions [1.25]](https://trello.com/c/CJVJPmwq/11-175-issues-creating-new-promotions-125)

Jerry77oz (approved):

PR #1827, opened 2018-03-19
Added the ability to use the Google Closure Compiler

dmyroniuk:
BroadleafCommerce/QA#3232
Added the ability to use the Google Closure Compiler or the YUICompressor to minify the javascript resources.

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/config/BroadleafCommonConfig.java
+++ b/common/src/main/java/org/broadleafcommerce/common/config/BroadleafCommonConfig.java
@@ -44,11 +44,11 @@ import org.springframework.orm.jpa.vendor.HibernateJpaVendorAdapter;
 @Configuration
 public class BroadleafCommonConfig {
 
-    @Value("${minify.compilerOptions.compilerLanguageIn:ECMASCRIPT_2017}")
+    @Value("${minify.closure.compiler.languageIn:ECMASCRIPT_2017}")
     protected String compilerLanguageIn;
-    @Value("${minify.compilerOptions.compilerLanguageOut:ECMASCRIPT5}")
+    @Value("${minify.closure.compiler.languageOut:ECMASCRIPT5}")
     protected String compilerLanguageOut;
-    @Value("${minify.compilerOptions.warningLevel:QUIET}")
+    @Value("${minify.closure.compiler.warningLevel:QUIET}")
     protected String warningLevel;
 
     /**
--- a/common/src/main/java/org/broadleafcommerce/common/config/BroadleafCommonConfig.java
+++ b/common/src/main/java/org/broadleafcommerce/common/config/BroadleafCommonConfig.java
@@ -20,9 +20,19 @@
  */
 package org.broadleafcommerce.common.config;
 
+import com.google.javascript.jscomp.ClosureCodingConvention;
+import com.google.javascript.jscomp.CompilationLevel;
+import com.google.javascript.jscomp.Compiler;
+import com.google.javascript.jscomp.CompilerOptions;
+import com.google.javascript.jscomp.WarningLevel;
+import com.google.javascript.jscomp.parsing.parser.FeatureSet;
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.beans.factory.config.ConfigurableBeanFactory;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Scope;
 import org.springframework.orm.jpa.JpaVendorAdapter;
 import org.springframework.orm.jpa.vendor.HibernateJpaVendorAdapter;
 
@@ -33,6 +43,13 @@ import org.springframework.orm.jpa.vendor.HibernateJpaVendorAdapter;
  */
 @Configuration
 public class BroadleafCommonConfig {
+
+    @Value("${minify.compilerOptions.compilerLanguageIn:ECMASCRIPT_2017}")
+    protected String compilerLanguageIn;
+    @Value("${minify.compilerOptions.compilerLanguageOut:ECMASCRIPT5}")
+    protected String compilerLanguageOut;
+    @Value("${minify.compilerOptions.warningLevel:QUIET}")
+    protected String warningLevel;
 
     /**
      * Other enterprise/mulititenant modules override this adapter to provide one that supports dynamic filtration
@@ -46,6 +63,66 @@ public class BroadleafCommonConfig {
         // for read only transactions. When we advance Hibernate, we should look at not blocking Spring's connection preparation.
         vendorAdapter.setPrepareConnection(false);
         return vendorAdapter;
+    }
+
+    @Bean
+    @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
+    public Compiler blClosureCompiler() {
+        return new Compiler();
+    }
+
+    @Bean
+    public CompilerOptions blClosureCompilerOptions() {
+        CompilerOptions options = new CompilerOptions();
+
+        options.setCodingConvention(new ClosureCodingConvention());
+
+        if (StringUtils.isNoneBlank(compilerLanguageIn)) {
+            CompilerOptions.LanguageMode languageIn = CompilerOptions.LanguageMode.valueOf(compilerLanguageIn);
+            if (StringUtils.isBlank(compilerLanguageOut)) {
+                options.setLanguage(languageIn);
+            } else {
+                options.setLanguageIn(languageIn);
+            }
+        }
+
+        if(StringUtils.isNoneBlank(compilerLanguageOut)) {
+            options.setLanguageOut(CompilerOptions.LanguageMode.valueOf(compilerLanguageOut));
+        }
+
+        CompilationLevel.SIMPLE_OPTIMIZATIONS.setOptionsForCompilationLevel(options);
+
+        WarningLevel.valueOf(warningLevel).setOptionsForWarningLevel(options);
+
+        options.setClosurePass(true);
+        options.setRewritePolyfills(toFeatureSet(options.getLanguageIn()).contains(FeatureSet.ES6));
+        options.setStrictModeInput(true);
+
+        return options;
+    }
+
+    protected FeatureSet toFeatureSet(CompilerOptions.LanguageMode languageMode) {
+        switch (languageMode) {
+            case ECMASCRIPT3:
+                return FeatureSet.ES3;
+            case ECMASCRIPT5:
+            case ECMASCRIPT5_STRICT:
+                return FeatureSet.ES5;
+            case ECMASCRIPT_2015:
+                return FeatureSet.ES6_MODULES;
+            case ECMASCRIPT_2016:
+                return FeatureSet.ES7_MODULES;
+            case ECMASCRIPT_2017:
+                return FeatureSet.ES8_MODULES;
+            case ECMASCRIPT_NEXT:
+                return FeatureSet.ES_NEXT;
+            case ECMASCRIPT6_TYPED:
+                return FeatureSet.TYPESCRIPT;
+            case NO_TRANSPILE:
+                throw new IllegalStateException();
+        }
+
+        throw new IllegalStateException();
     }
 
 }
--- a/common/src/main/java/org/broadleafcommerce/common/config/BroadleafCommonConfig.java
+++ b/common/src/main/java/org/broadleafcommerce/common/config/BroadleafCommonConfig.java
@@ -18,21 +18,12 @@
 /**
  * 
  */
+
 package org.broadleafcommerce.common.config;
 
-import com.google.javascript.jscomp.ClosureCodingConvention;
-import com.google.javascript.jscomp.CompilationLevel;
-import com.google.javascript.jscomp.Compiler;
-import com.google.javascript.jscomp.CompilerOptions;
-import com.google.javascript.jscomp.WarningLevel;
-import com.google.javascript.jscomp.parsing.parser.FeatureSet;
-import org.apache.commons.lang3.StringUtils;
-import org.springframework.beans.factory.annotation.Value;
-import org.springframework.beans.factory.config.ConfigurableBeanFactory;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.context.annotation.Scope;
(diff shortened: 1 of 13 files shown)


common/src/main/java/org/broadleafcommerce/common/config/BroadleafCommonConfig.java
```diff
@@ -48,4 +65,64 @@ public JpaVendorAdapter blJpaVendorAdapter() {
...
+                return FeatureSet.ES_NEXT;
+            case ECMASCRIPT6_TYPED:
+                return FeatureSet.TYPESCRIPT;
+            case NO_TRANSPILE:
+                throw new IllegalStateException();
+        }
+
+        throw new IllegalStateException();
+    }
+
```
cja769: All of this should be in an `@Conditional` on a closure-compiler class existing since we're making it an optional dependency.

common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceMinificationServiceImpl.java
```diff
@@ -64,20 +80,45 @@
     @Value("${minify.disableOptimizations}")
     protected boolean disableOptimizations;
 
+    @Value("${minify.closure.compiler.enabled}")
+    protected boolean closureCompilerEnabled;
+
+    protected CompilerOptions compilerOptions;
+
+    protected SystemPropertiesService systemPropertiesService;
+
```
cja769: This should be `protected Environment environment`. Our system properties are added to the Spring environment so for consistency we've been moving away from injecting the `SystemPropertiesService` in favor of using `Environment`

common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceMinificationServiceImpl.java
```diff
@@ -64,20 +80,45 @@
     @Value("${minify.disableOptimizations}")
     protected boolean disableOptimizations;
 
+    @Value("${minify.closure.compiler.enabled}")
+    protected boolean closureCompilerEnabled;
```
cja769: Having a boolean to tell a service to do one of two very different implementations is a dead giveaway that we need to have 2 implementations of `ResourceMinificationService`. One that implements how we do it for YUICompressor and one for Google Closure Compiler. We can either create the bean `blResourceMinificationService` in a configuration file based on if Google Closure Compiler is on the classpath or not or we can add that `@Conditional` directly on the class an let component scanning dictate which one is created. If there's common logic between the two minification service then an Abstract class should be created that has the common logic implemented

common/src/main/java/org/broadleafcommerce/common/config/BroadleafCommonConfig.java
```diff
@@ -48,4 +65,64 @@ public JpaVendorAdapter blJpaVendorAdapter() {
         return vendorAdapter;
     }
 
+    @Bean
+    @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
+    public Compiler blClosureCompiler() {
+        return new Compiler();
+    }
+
```
cja769: I'm not really sure why this is a prototype bean. It seems that we only need one implementation.

common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceMinificationService.java
```diff
@@ -20,24 +20,43 @@
...
+ * "ECMASCRIPT6_TYPED (experimental), ECMASCRIPT_2015, ECMASCRIPT_2016, "
+ * "ECMASCRIPT_2017, ECMASCRIPT_NEXT</li>
+ *
+ *  <li>minify.closure.compiler.languageOut - Sets ECMAScript version to use for the output.
+ *  Options: ECMASCRIPT3, ECMASCRIPT5, ECMASCRIPT5_STRICT, "
+ * "ECMASCRIPT6_TYPED (experimental), ECMASCRIPT_2015, ECMASCRIPT_2016, "
+ * "ECMASCRIPT_2017, ECMASCRIPT_NEXT, NO_TRANSPILE</li>
+ *
+ *  <li>minify.closure.compiler.warningLevel - Warnings level. Possible values: QUIET, DEFAULT, VERBOSE</li>
+ * </ul>
```
cja769: Since we're creating multiple implementations of this class it's probably better to put the docs that are relevant to that implementation there instead of in the interface. The interface should have minification library agnostic information about what this interface is used for and what the methods do.

common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceMinificationService.java
```diff
@@ -56,9 +75,11 @@
      * Indicates whether or not the system is allowed to attempt to minify individual files. This can be useful if
      * the YUI compressor is failing to minify JavaScript/CSS due to syntax errors and you are attempting to track
      * down which file is problematic. It should not be enabled in a production environment.
-     * 
+     *
+     * @deprecated
      * @return the value of the system property "minify.allowSingleMinification"
      */
+    @Deprecated
     boolean getAllowSingleMinification();
```
cja769, line 58: We probably need to make sure that it's fine but I would rather have this method just moved to the YUI compressor implementation. I don't know why this was on the interface in the first place.

common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceMinificationServiceImpl.java
```diff
@@ -64,20 +80,45 @@
...
+    @Qualifier("blSystemPropertiesService")
+    public void setSystemPropertiesService(SystemPropertiesService systemPropertiesService) {
+        this.systemPropertiesService = systemPropertiesService;
+    }
+
+    @Lookup
+    protected Compiler getClosureCompiler() {
+        return null;
+    }
+
```
cja769: I'm guessing this is the reason on why the `Compiler` was a prototype bean. I'm not really sure what advantage making the `Compiler` bean a prototype is with using `@Lookup` instead of simply creating a singleton bean and injecting it is.

pom.xml
```diff
@@ -1591,7 +1592,12 @@
             <dependency>
                 <groupId>com.yahoo.platform.yui</groupId>
                 <artifactId>yuicompressor</artifactId>
-                <version>2.4.7</version>
+                <version>2.4.8</version>
+            </dependency>
+            <dependency>
+                <groupId>com.google.javascript</groupId>
+                <artifactId>closure-compiler</artifactId>
+                <version>${closure.compiler.version}</version>
```
cja769, line 1605: I think this dependency and/or the YUI compressor should be optional dependency. I'm leaning towards just the closure-compiler being optional. This will give us a way to use @Conditionals to determine if we're using the closure-compiler to minify JS.

cja769 (dismissed):
There's a couple open-ended questions. One is I don't know the intent of 5.3. I don't know if it's getting merged into 5.4 and live like the rest of Broadleaf or if this is completely separate from the mainline. The reason this is important is because I've heard of multiple issues with YUICompressor with other clients so I think changing this is good to do. Second, I know that O'Reilly replaced YUICompressor in their project and they're targeting Java 7 so it leads my to believe that there's a library out there that's Java 7 compatible which would allow us to put this in 5.2. Third, I don't know if we're planning on fully replacing YUICompressor or having it live side-by-side with a new library. If we're replacing it then we need to make sure that it's indicated that it's deprecated (not that many clients are even aware of what library we even use for minification). 

Either way the main thing that needs to happen is to have 2 separate implementations of the ResourceMinificationService where the bean is created based on the existence of the google closure dependency. I think it's best to make the google closure dependency optional since the YUICompressor dependency is still needed for CSS compression therefore it'll always be on the classpath anyways so technically a client changing to using the google compressor would be optional.

common/src/main/java/org/broadleafcommerce/common/config/BroadleafCommonConfig.java
cja769: Sorry if the code snippet is vague. I'm referring to all of the Google Closure configuration

cja769:
I contacted O'Reilly to ask which library they replaced YUICompressor with and they actually replaced it with a Javascript build tool and just turned off Broadleaf's minification process. Additionally I talked to Phillip Verhyden about this and he confirmed what I was thinking. All of the 5.x line should be Java 7 compatible so we can't add the Google library since it requires Java 8.

common/src/main/java/org/broadleafcommerce/common/resource/service/GoogleClosureJavascriptMinificationServiceImpl.java
```diff
@@ -0,0 +1,134 @@
...
+ *  Options: ECMASCRIPT3, ECMASCRIPT5, ECMASCRIPT5_STRICT, "
+ * "ECMASCRIPT6_TYPED (experimental), ECMASCRIPT_2015, ECMASCRIPT_2016, "
+ * "ECMASCRIPT_2017, ECMASCRIPT_NEXT</li>
+ *
+ *  <li>minify.closure.compiler.languageOut - Sets ECMAScript version to use for the output.
+ *  Options: ECMASCRIPT3, ECMASCRIPT5, ECMASCRIPT5_STRICT, "
+ * "ECMASCRIPT6_TYPED (experimental), ECMASCRIPT_2015, ECMASCRIPT_2016, "
+ * "ECMASCRIPT_2017, ECMASCRIPT_NEXT, NO_TRANSPILE</li>
+ *
+ *  <li>minify.closure.compiler.warningLevel - Warnings level. Possible values: QUIET, DEFAULT, VERBOSE</li>
```
phillipuniverse: If you are documenting this, you should also include the default values for all of these that you have set.

common/src/main/java/org/broadleafcommerce/common/resource/service/JavascriptMinificationService.java
```diff
@@ -0,0 +1,26 @@
...
+ * Alternatively, the Commercial License may be replaced with a mutually agreed upon license (the "Custom License")
+ * between you and Broadleaf Commerce. You may not use this file except in compliance with the applicable license.
+ * #L%
+ */
+package org.broadleafcommerce.common.resource.service;
+
+import java.io.Reader;
+import java.io.Writer;
+
+public interface JavascriptMinificationService {
```
phillipuniverse: Can you `@author` and give a little something about this in a Javadoc like a link to look at YUI or Google Closure or something? just looks weird with no javadoc on this

common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceMinificationServiceImpl.java
```diff
@@ -107,44 +100,46 @@ public Resource minify(Resource originalResource) {
...
         } catch (Exception e) {
-            LOG.warn("Could not minify resources, returned unminified bytes", e);
+            LOG.warn(COULD_NOT_MINIFY_RESOURCES_RETURNED_UNMINIFIED_BYTES, e);
             return originalResource;
         }
-        
-        return new GeneratedResource(minifiedBytes, filename);
     }
     
+    // TODO Change method signature to throw a ResourceMinificationException in 5.4
```
phillipuniverse, line 126: I don't think that we should change this to throw an exception ever; we should just straight up handle it  right?

common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceMinificationServiceImpl.java
```diff
@@ -107,44 +100,46 @@ public Resource minify(Resource originalResource) {
...
+                jsMinificationService.minifyJs(filename, in, out);
+            } else if (CSS_TYPE.equals(type)) {
+                cssMinificationService.minifyCss(filename, in, out);
+            } else {
+                throw new ResourceMinificationException("Minification is not supported for file: " + filename);
+            }
+        } catch (ResourceMinificationException e) {
+            // We're throwing and catching ResourceMinificationException for backwards compatibility. 
+            // In 5.4 this method will throw a ResourceMinificationException
+            throw new RuntimeException(e);
```
phillipuniverse: Just throw `IOException(e)` instead, somebody else could check the cause

common/src/main/java/org/broadleafcommerce/common/resource/service/YUIJavscriptionMinificationServiceImpl.java
```diff
@@ -0,0 +1,116 @@
...
+ *  <li>minify.munge - if true, will replace variable names with shorter versions</li>
+ *  <li>minify.verbose - if true, will display extra logging information to the console</li>
+ *  <li>minify.preserveAllSemiColons - if true, will never remove semi-colons, even if two in a row exist</li>
+ *  <li>minify.disableOptimizations - if true, will disable some micro-optimizations that are performed</li>
+ * </ul>
+ * 
+ **/
+@Service("blJavascriptMinificationService")
+@ConditionalOnMissingClass(value = "com.google.javascript.jscomp.Compiler")
+public class YUIJavscriptionMinificationServiceImpl implements JavascriptMinificationService {
```
phillipuniverse: This is a typo

common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceMinificationServiceImpl.java
```diff
@@ -107,44 +100,46 @@ public Resource minify(Resource originalResource) {
...
-            LOG.trace("Minification is disabled, returning original resource");
+            LOG.trace(MINIFICATION_IS_DISABLED_RETURNING_ORIGINAL_RESOURCE);
             return originalResource;
         }
-        
+
         String type = getFileType(originalResource, filename);
         if (type == null) {
-            LOG.info("Unsupported minification resource: " + filename);
+            LOG.trace("Unsupported minification resource: " + filename);
```
phillipuniverse: Why not continue to `LOG.info` here?

phillipuniverse (changes requested):

common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceMinificationServiceImpl.java
cja769: We handle it in the call that calls this method by just returning the original file. Technically we could feed the BufferedReader into the BufferedWriter but that seems pretty round about when the calling method can just return the original file.

common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceMinificationServiceImpl.java
cja769: Info is fine. I didn't realize that was the original logging level. Since ReferenceSite has info logging on it seemed extra loud since this method gets called for image resources as well as JS, and CSS

phillipuniverse (approved):

phillipuniverse:
:shipit:

PR #1831, opened 2018-03-22
Allow image uploads in the admin to be limited separately from file uploads

jfleschler:


phillipuniverse:
@jfleschler @danielcolgrove you can't just change what the property is called in a non-backwards compatible way in a patch release, and even in a major/minor release it should at least be documented. This change breaks everyone that has been relying on the old property name.

phillipuniverse:
Oh yeah, and since the check is in bytes the value of `ONE_MEGABYTE` is actually 1kb.

phillipuniverse:
Fixed in my commit above.

PR #1834, opened 2018-03-27
Changed length check for string in setValue() method.

Vitalij-Voronkoff:
BroadleafCommerce/QA#3261

Requires:BroadleafCommerce/Theme#136


charchar (approved):
Looks good

PR #1843, opened 2018-04-05
Fixed bug where hidden modals on the page prevented clearing entity form state on submission

cja769:
In the success handler when submitting an entity form via ajax we were preventing the entity form change state from being cleared if there were any modals on the page. The intent was to not clear state of the underlying entity form if a modal form was being submitted. However, by simply checking for any modals on the page, we prevented the normal behavior when there are benign, hidden modals on the page. This changes now makes it so that we're specifically checking that the form being submitted is contained by a modal which is more definitive that we aren't submitting the main entity form for the page.


phillipuniverse:
LGTM 🎉

PR #1849, opened 2018-04-12
Address issue where entities where getting incorrectly their updated date changed when browsing the site

Jerry77oz:
BroadleafCommerce/QA#3283

This is meant to address the first point in order to get the client past the issue.

- Prevent `AdminAuditableListener` from running on site


danielcolgrove (approved):

jefffischer (approved):
Seems ok

Issue #798, opened 2014-04-03
not supported image type with cmsstatic image processing


mohitxaton:
find details at 

http://forum.broadleafcommerce.org/viewtopic.php?f=14&t=2774

blc version 3.0.5-GA

PR #1854, opened 2018-04-16
Replaced MimeUtil with Apache Tika for mime type detection

danielcolgrove:
MimeUtil is no longer supported and incorrectly detects MP4 files as application/octet-stream.  Updating to use Apache Tika to detect mime types.  Fixes https://github.com/BroadleafCommerce/QA/issues/3290

### DIFF
--- a/admin/broadleaf-contentmanagement-module/src/main/java/org/broadleafcommerce/cms/file/service/StaticAssetServiceImpl.java
+++ b/admin/broadleaf-contentmanagement-module/src/main/java/org/broadleafcommerce/cms/file/service/StaticAssetServiceImpl.java
@@ -2,6 +2,22 @@
  * #%L
  * BroadleafCommerce CMS Module
  * %%
+ * Copyright (C) 2009 - 2018 Broadleaf Commerce
+ * %%
+ * Licensed under the Broadleaf Fair Use License Agreement, Version 1.0
+ * (the "Fair Use License" located  at http://license.broadleafcommerce.org/fair_use_license-1.0.txt)
+ * unless the restrictions on use therein are violated and require payment to Broadleaf in which case
+ * the Broadleaf End User License Agreement (EULA), Version 1.1
+ * (the "Commercial License" located at http://license.broadleafcommerce.org/commercial_license-1.1.txt)
+ * shall apply.
+ * 
+ * Alternatively, the Commercial License may be replaced with a mutually agreed upon license (the "Custom License")
+ * between you and Broadleaf Commerce. You may not use this file except in compliance with the applicable license.
+ * #L%
+ */
+/*
+ * BroadleafCommerce CMS Module
+ * %%
  * Copyright (C) 2009 - 2016 Broadleaf Commerce
  * %%
  * Licensed under the Broadleaf Fair Use License Agreement, Version 1.0
@@ -19,6 +35,7 @@ package org.broadleafcommerce.cms.file.service;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.tika.Tika;
 import org.broadleafcommerce.cms.field.type.StorageType;
 import org.broadleafcommerce.cms.file.StaticAssetMultiTenantExtensionManager;
 import org.broadleafcommerce.cms.file.dao.StaticAssetDao;
@@ -48,11 +65,6 @@ import java.util.Map;
 import java.util.Random;
 
 import javax.annotation.Resource;
-
-import eu.medsea.mimeutil.MimeType;
-import eu.medsea.mimeutil.MimeUtil;
-import eu.medsea.mimeutil.detector.ExtensionMimeDetector;
-import eu.medsea.mimeutil.detector.MagicMimeMimeDetector;
 
 /**
  * Created by bpolster.
@@ -99,11 +111,6 @@ public class StaticAssetServiceImpl implements StaticAssetService {
     @Override
     public Long findTotalStaticAssetCount() {
         return staticAssetDao.readTotalStaticAssetCount();
-    }
-
-    static {
-        MimeUtil.registerMimeDetector(ExtensionMimeDetector.class.getName());
-        MimeUtil.registerMimeDetector(MagicMimeMimeDetector.class.getName());
     }
 
     protected String getFileExtension(String fileName) {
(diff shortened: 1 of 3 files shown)


phillipuniverse:
@danielcolgrove you had removed the Twelvemonkeys library, I added it back (or more technically, removed it from your commit). That library actually doesn't have anything to do with image _detection_ but rather image _transformation_. It was originally added in #798 and handles transforming CMYK JPEg images. The reason that you would not see it anywhere in our codebase outside of this dependency, is that it registers itself via jar manifest files into the `ImageIO` library.

The place this is used is Broadleaf is [here](https://github.com/BroadleafCommerce/BroadleafCommerce/blob/develop-5.2.x/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/artifact/image/ImageArtifactProcessor.java#L75). If we want to remove this library with no other consequences we would need to:

1. Find some other `ImageIO` plugin that supports CMYK JPEG images
2. Use something other than `ImageIO` to perform our image transformations (maybe Tika can do that?)

Either way, (1) or (2) should be a separate change, unrelated to this one.

phillipuniverse:
Found a fun little issue with this, manifests itself in BroadleafCommerce/OMS#346. Here is the situation:

- OMS has updated its dependency to use `broadleaf-framework:5.2.4-SNAPSHOT`
- This update transitively brings in `broadleaf-common:5.2.4-SNAPSHOT`
- Even though the `MimeUtil` was only used in the CMS module, we had it declared as a dependency in common
- OMS also depends on `broadleaf-enterprise:3.2.0-GA` for its CSR functionality
- `broadleaf-enterprise:3.2.0-GA` transitively depends on `broadleaf-content-management-module:5.2.0-GA`
- Since nothing else in OMS has higher precedence in transitive dependencies, it resolves `broadleaf-content-management-module:5.2.0-GA`

So now we are in this situation in OMS at compile time:

- `broadleaf-common:5.2.4-SNAPSHOT` with no mimeutil transitive dependency declared and thus the dependency never resolves
- `broadleaf-content-management-module:5.2.0-GA` that requires Mimeutil in the `StaticAssetService`

This causes the following exception when running the integration tests in OMS which starts up the application context:

```console
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'blStaticAssetService' defined in URL [jar:file:/Users/phillip/.m2/repository/org/broadleafcommerce/broadleaf-contentmanagement-module/5.2.0-GA/broadleaf-contentmanagement-module-5.2.0-GA.jar!/org/broadleafcommerce/cms/file/service/StaticAssetServiceImpl.class]: Instantiation of bean failed; nested exception is java.lang.NoClassDefFoundError: Could not initialize class org.broadleafcommerce.cms.file.service.StaticAssetServiceImpl
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.instantiateBean(AbstractAutowireCapableBeanFactory.java:1159)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1103)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:513)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:483)
	at org.springframework.beans.factory.support.AbstractBeanFactory$1.getObject(AbstractBeanFactory.java:312)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:230)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:308)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:202)
	at org.springframework.context.annotation.CommonAnnotationBeanPostProcessor.autowireResource(CommonAnnotationBeanPostProcessor.java:522)
	at org.springframework.context.annotation.CommonAnnotationBeanPostProcessor.getResource(CommonAnnotationBeanPostProcessor.java:496)
	at org.springframework.context.annotation.CommonAnnotationBeanPostProcessor$ResourceElement.getResourceToInject(CommonAnnotationBeanPostProcessor.java:627)
	at org.springframework.beans.factory.annotation.InjectionMetadata$InjectedElement.inject(InjectionMetadata.java:169)
	at org.springframework.beans.factory.annotation.InjectionMetadata.inject(InjectionMetadata.java:88)
	at org.springframework.context.annotation.CommonAnnotationBeanPostProcessor.postProcessPropertyValues(CommonAnnotationBeanPostProcessor.java:318)
	... 140 more
Caused by: java.lang.NoClassDefFoundError: Could not initialize class org.broadleafcommerce.cms.file.service.StaticAssetServiceImpl
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.__newInstance(DelegatingConstructorAccessorImpl.java:45)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java)
```

In order to resolve this, I think we need to put the mimeutil dependency back into common as a transitive dependency, and learn from our previous mistake and move the Tika library into only the CMS (which is the only place it is used). There is another solution that involves manually setting the version for the `broadleaf-content-management-module` to `5.2.4-SNAPSHOT` but this seems really wrong. OMS doesn't explicitly need that and thus should not need to depend on it; this is something that is set by enterprise that enterprise requires.

A remediation strategy for when we can confidently remove the Mimeutil library is when:
- There are no other Broadleaf modules that depend on CMS < 5.2.4-GA (directly _or_ transitively)
- We manually harmonize all the Broadleaf framework versions to be the same via a `<dependencyManagement>` section in any modules that depend on the framework. Basically a BOM but solely for the Broadleaf framework. This way we would not run into conflicting transitive versions 

Also important to note that this would not normally affect anything other than our own framework and module dependencies because we manually set every version of all Broadleaf artifacts directly in each implementation project pom (through the BOM or parent poms).

PR #1856, opened 2018-04-19
Fixed the issue where adding inactive Skus to cart causes error

riteshadhikari17:
https://github.com/BroadleafCommerce/QA/issues/3244


core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/workflow/add/ValidateAddRequestActivity.java
```diff
@@ -201,15 +207,16 @@ protected Sku findMatchingSku(Product product, Map<String, String> attributeValu
...
+    protected void addSkuToCart(Sku sku, OrderItemRequestDTO orderItemRequestDTO, Product product, CartOperationRequest request) throws InventoryUnavailableException{
         // If we couldn't find a sku, then we're unable to add to cart.
         if (!hasSkuOrIsNonDiscreteOI(sku, orderItemRequestDTO)) {
             handleIfNoSku(orderItemRequestDTO, product);
         } else if (sku == null) {
             handleIfNonDiscreteOI(orderItemRequestDTO);
         } else if (!sku.isActive()) {
-            throw new IllegalArgumentException("The requested skuId (" + sku.getId() + ") is no longer active");
-        } else {
+            throw new InventoryUnavailableException("Sku is not Active");
```
cja769: Saying that a Sku is inactive means that it's not for sale whereas the InventoryUnavailableException is really used for when a Sku is active but it doesn't have inventory at this time. It's a subtle difference but if we're going to change the signature to return something that's not an IllegalArgumentException then I think we really should create a new Exception. IMO there's a reason we're throwing an IllegalArgumentException. It's because there's no other Exception in Broadleaf that makes sense therefore we should make one

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/workflow/add/ValidateAddRequestActivity.java
```diff
@@ -132,12 +139,11 @@ protected Sku determineSku(Product product, Long skuId, Map<String, String> attr
         if (sku == null && product != null) {
             // Set to the default sku
             if (cannotSellDefaultSku(product)) {
-                throw new RequiredAttributeNotProvidedException("Unable to find non-default sku matching given options and cannot sell default sku", null);
+                throw new InventoryUnavailableException("Sku is not Available");
```
cja769: Yeah don't change this. RequiredAttributeNotProvidedException is exactly what needs to be thrown here. It indicates that the you didn't send valid product option values

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/workflow/add/ValidateAddRequestActivity.java
```diff
@@ -82,10 +84,15 @@
...
-            Sku sku = determineSku(product, orderItemRequestDTO.getSkuId(), orderItemRequestDTO.getItemAttributes(),
+            Sku sku;
+            try {
+                sku = determineSku(product, orderItemRequestDTO.getSkuId(), orderItemRequestDTO.getItemAttributes(),
                                    (ActivityMessages) context);
+                addSkuToCart(sku, orderItemRequestDTO, product, request);
 
-            addSkuToCart(sku, orderItemRequestDTO, product, request);
+            }catch(InventoryUnavailableException e){
+                throw e;
```
cja769: No reason to catch an exception and then immediately throw it with no changes

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/catalog/dao/ProductOptionDaoImpl.java
```diff
@@ -17,17 +17,11 @@
...
-import org.broadleafcommerce.core.catalog.domain.Product;
-import org.broadleafcommerce.core.catalog.domain.ProductImpl;
-import org.broadleafcommerce.core.catalog.domain.ProductOption;
-import org.broadleafcommerce.core.catalog.domain.ProductOptionImpl;
-import org.broadleafcommerce.core.catalog.domain.ProductOptionValue;
-import org.broadleafcommerce.core.catalog.domain.ProductOptionValueImpl;
-import org.broadleafcommerce.core.catalog.domain.ProductOptionXref;
-import org.broadleafcommerce.core.catalog.domain.ProductOptionXrefImpl;
-import org.broadleafcommerce.core.catalog.domain.SkuProductOptionValueXrefImpl;
+import org.broadleafcommerce.core.catalog.domain.*;
```
cja769: In our code style we don't use the wildcard import

cja769 (changes requested):
I'm not actually sure what the best way to back port the ProductOptionDaoImpl changes would be. Generally you'd want to cherry pick the commits but the changes you have included include changes from multiple commits. I'm not familiar with what's the correct path on this is. @danielcolgrove what do you think the best thing to do here is?

riteshadhikari17:
With current design, it is not possible to hide unavailable skus because black small might not be available but red or silver small is. So small option needs to be present. I'll introduce two new exceptions SkuNotFoundException and SkuNotActiveException. How should it be handled at the front end? Regardless of underlying cause its not available for consumer so there must be some message stating the chosen item cannot be purchased. Is it okay to show same message as the inventory unavailable?

cja769:
The pieces of this PR that needed to be merge was made by BroadleafCommerce/BroadleafCommerce#1859

PR #1859, opened 2018-04-23
Disallow adding archived skus to cart

cja769:
Backport ProductOptionDaoImpl changes so that archived additional skus aren't returned when trying to find skus that match a product option name/value pair

BroadleafCommerce/QA#3244


cja769:
The main change added around filtering Archived Skus from the result set was done in 5.1.x but never ported to 5.0.x

This is the originally QA issue created to fix this issue https://github.com/BroadleafCommerce/QA/issues/2825

PR #1860, opened 2018-04-23
Implemented SkuNotActiveException and SkuNotAvailableException

riteshadhikari17:
https://github.com/BroadleafCommerce/QA/issues/3244

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/exception/SkuNotActiveException.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/exception/SkuNotActiveException.java
@@ -1,0 +1,44 @@
+/*
+ * #%L
+ * BroadleafCommerce Framework
+ * %%
+ * Copyright (C) 2009 - 2016 Broadleaf Commerce
+ * %%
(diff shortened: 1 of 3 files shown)


cja769:
As Daniel mentioned in BroadleafCommerce/QA#3244 we don't want to change the exceptions thrown because clients likely are catching those exceptions and have messaging specifically for those situations therefore we don't want to change something like this in a patch release. Generally we try not to affect clients with changes in patch releases so that they can update without having any affects.

Additionally, like I had said on a different PR, we don't want to change the `RequiredAttributeNotProvidedException` to anything else because that exception explains what's happening. Generally the sku couldn't be found because the product options are invalid. In the case where the product options are "valid" and the sku associated is archived then it's a bit of an edge case but still, technically, the attributes provided were invalid so the exception is still relevant.

riteshadhikari17:
I disagree with this statement: "Generally the sku couldn't be found because the product options are invalid." because all the required attributes are still valid for the product but that specific sku is not available. From consumer's point he picked a color and a size (both are valid attributes for product) but still an error he is receiving is "All options are required". How is that correct? Sku was not found because it has been taken off not because the attributes were invalid.

PR #1874, opened 2018-05-03
Enable creating SystemEvents without a dependency on the ScheduledJobsAndEvents module

cja769:
Created a new BroadleafApplicationEvent that can be created a published via the ApplicationContext. This event is used to indirectly create SystemEvents from the ScheduledJobsAndEvents module without having a direct dependency on the module itself.


phillipuniverse:
Required to resolve BroadleafCommerce/QA#3172.

Issue #404, opened 2013-09-25
IsPresent rule builder operator does not work for custom field


jefffischer:
[ERROR] 10:42:22 TransactionInterceptor - Application exception overridden by commit exception
org.broadleafcommerce.common.exception.ServiceException: Problem updating entity : java.lang.RuntimeException: org.broadleafcommerce.openadmin.web.rulebuilder.MVELT
ranslationException: Cannot format value for the field (getCustomerAttributes()---THING) based on field type. The type of field is Integer, and you entered: (null)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:525)
        at org.broadleafcommerce.openadmin.server.service.DynamicEntityRemoteService.recreateSpecificServiceException(DynamicEntityRemoteService.java:128)
        ...
Caused by: org.broadleafcommerce.openadmin.server.service.persistence.PersistenceException: java.lang.RuntimeException: org.broadleafcommerce.openadmin.web.rulebuilder.MVELTranslationException: Cannot format value for the field (getCustomerAttributes()---THING) based on field type. The type of field is Integer, and you entered: (null)
        at org.broadleafcommerce.openadmin.server.service.persistence.module.provider.RuleFieldPersistenceProvider.populateValue(RuleFieldPersistenceProvider.java:161)
        at org.broadleafcommerce.openadmin.server.service.persistence.module.BasicPersistenceModule.createPopulatedInstance(BasicPersistenceModule.java:277)
        at org.broadleafcommerce.openadmin.server.service.persistence.module.BasicPersistenceModule.update(BasicPersistenceModule.java:646)
        at org.broadleafcommerce.openadmin.server.service.persistence.module.BasicPersistenceModule.update(BasicPersistenceModule.java:818)
        at org.broadleafcommerce.openadmin.server.service.persistence.PersistenceManagerImpl.update(PersistenceManagerImpl.java:318)
        at com.broadleafcommerce.tenant.service.MultiTenantPersistenceManagerImpl.update(MultiTenantPersistenceManagerImpl.java:73)
[...]

PR #1875, opened 2018-05-03
Fixed issue where the query cache wasn't invalidated when OMS updated sku inventory

cja769:
This change fixes an issue where when a sku with CHECK_INVENTORY was canceled or returned via OMS and the project wasn't running AdvancedInventory, the site node query caches weren't being invalidated. This utilize Spring events to publish a cache invalidation event that is picked up if ScheduledJobsAndEvents is included in the project.

This PR depends on
https://github.com/BroadleafCommerce/BroadleafCommerce/pull/1874
https://github.com/BroadleafCommerce/ScheduledJobsAndEvents/pull/80

This PR should be rebased after https://github.com/BroadleafCommerce/BroadleafCommerce/pull/1874 is merged

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/inventory/service/InventoryServiceImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/inventory/service/InventoryServiceImpl.java
@@ -20,8 +20,10 @@ package org.broadleafcommerce.core.inventory.service;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.broadleafcommerce.common.event.BroadleafJobsEvent;
-import org.broadleafcommerce.common.event.BroadleafJobsEventDetail;
+import org.broadleafcommerce.common.event.BroadleafSystemEvent;
+import org.broadleafcommerce.common.event.BroadleafSystemEvent.BroadleafEventScopeType;
+import org.broadleafcommerce.common.event.BroadleafSystemEvent.BroadleafEventWorkerType;
+import org.broadleafcommerce.common.event.BroadleafSystemEventDetail;
 import org.broadleafcommerce.common.extension.ExtensionResultHolder;
 import org.broadleafcommerce.common.extension.ExtensionResultStatusType;
 import org.broadleafcommerce.common.util.TransactionUtils;
@@ -332,11 +334,11 @@ public class InventoryServiceImpl implements ContextualInventoryService {
     }
 
     protected void invalidateSkuInventoryNonTransactional(String className, String id) {
-        Map<String, BroadleafJobsEventDetail> detailMap = new HashMap<>();
-        detailMap.put("CACHE_REGION", new BroadleafJobsEventDetail("CACHE_REGION", "Cache Region", "blProducts"));
-        detailMap.put("ENTITY_TYPE", new BroadleafJobsEventDetail("ENTITY_TYPE", "Entity Type", className));
-        detailMap.put("IDENTIFIER", new BroadleafJobsEventDetail("IDENTIFIER", "Identifier", id));
-        BroadleafJobsEvent event = new BroadleafJobsEvent("CACHE", detailMap, "GLOBAL", "ANY", true);
+        Map<String, BroadleafSystemEventDetail> detailMap = new HashMap<>();
+        detailMap.put("CACHE_REGION", new BroadleafSystemEventDetail("Cache Region", "blProducts"));
+        detailMap.put("ENTITY_TYPE", new BroadleafSystemEventDetail("Entity Type", className));
+        detailMap.put("IDENTIFIER", new BroadleafSystemEventDetail("Identifier", id));
+        BroadleafSystemEvent event = new BroadleafSystemEvent("CACHE", detailMap, BroadleafEventScopeType.GLOBAL, BroadleafEventWorkerType.ANY, true);
         applicationContext.publishEvent(event);
     }
 
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/inventory/service/InventoryServiceImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/inventory/service/InventoryServiceImpl.java
@@ -20,6 +20,8 @@ package org.broadleafcommerce.core.inventory.service;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.broadleafcommerce.common.event.BroadleafJobsEvent;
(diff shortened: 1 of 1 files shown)


phillipuniverse:
Required to fully resolve BroadleafCommerce/QA#3172.

common/src/main/java/org/broadleafcommerce/common/event/BroadleafJobsEvent.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * A BroadleafApplicationEvent used so that we can communicate with the SystemJobsAndEvents module without having a dependency on it.
+ * By publishing a Spring Event with this detail, the SystemJobsAndEvents module will listen for this event and create a corresponding
+ * SystemEvent to be consumed. 
+ * 
+ * @author Jay Aisenbrey (cja769)
+ *
+ */
+public class BroadleafJobsEvent extends BroadleafApplicationEvent {
```
phillipuniverse: I think you should rename this to `BroadleafSystemEvent`, the `Jobs` part of this seems unnecessary.

common/src/main/java/org/broadleafcommerce/common/event/BroadleafJobsEventDetail.java
```diff
@@ -0,0 +1,98 @@
...
+import java.io.Serializable;
+
+/**
+ * Effectively a copy of com.broadleafcommerce.jobsevents.domain.dto.SystemEventDetailDTO
+ * to be used when creating a org.broadleafcommerce.common.event.BroadleafJobsEvent
+ * 
+ * @author Jay Aisenbrey (cja769)
+ *
+ */
+public class BroadleafJobsEventDetail implements Serializable {
```
phillipuniverse: Also would rename this guy to a `BroadleafSystemEventDetail`

common/src/main/java/org/broadleafcommerce/common/event/BroadleafJobsEventDetail.java
```diff
@@ -0,0 +1,98 @@
...
+
+    private static final long serialVersionUID = 1L;
+
+    protected String name;
+    protected String friendlyName;
+    protected String value;
+    protected String longValue;
```
phillipuniverse: I think the differnetiation between 'value' and 'longValue' is an unnecessary distinction. You should just have a `value` and parse it into the `longValue` or `value` when initializing the `SystemEvent`

common/src/main/java/org/broadleafcommerce/common/event/BroadleafJobsEventDetail.java
```diff
@@ -0,0 +1,98 @@
...
+ * to be used when creating a org.broadleafcommerce.common.event.BroadleafJobsEvent
+ * 
+ * @author Jay Aisenbrey (cja769)
+ *
+ */
+public class BroadleafJobsEventDetail implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    protected String name;
```
phillipuniverse: I think remove this property, just rely on the map key instead when creating the `SystemEventDetail`.

common/src/main/java/org/broadleafcommerce/common/event/BroadleafJobsEventDetail.java
phillipuniverse: I originally thought this was completely unnecessary but I think it _is_ worth having the `friendlyName` and `blob` as options. However, both of those fields should be completely optional. I expect that general usage would only set a `value`

common/src/main/java/org/broadleafcommerce/common/event/BroadleafJobsEvent.java
```diff
@@ -0,0 +1,99 @@
...
+    private static final long serialVersionUID = 1L;
+
+    protected Map<String, BroadleafJobsEventDetail> detailMap;
+    protected String type;
+
+    /**
+     * @see com.broadleafcommerce.jobsevents.service.type.EventScopeType
+     * Use the "type" field and not the "friendlyType" field
+     */
+    protected String scopeType;
```
phillipuniverse: It is worth making this an enum, probably sufficient to be a static inner enum within this class.

common/src/main/java/org/broadleafcommerce/common/event/BroadleafJobsEvent.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * @see com.broadleafcommerce.jobsevents.service.type.EventScopeType
+     * Use the "type" field and not the "friendlyType" field
+     */
+    protected String workerType;
```
phillipuniverse: It is worth making this an enum, probably sufficient to be a static inner enum within this class.

phillipuniverse (changes requested):

common/src/main/java/org/broadleafcommerce/common/event/BroadleafJobsEventDetail.java
cja769: How would I know if the original value was a longValue or not when creating the SystemEvent? Using `SystemEvent#setLongValue` has different operations than `SystemEvent#setValue` so it's an important distinction.

common/src/main/java/org/broadleafcommerce/common/event/BroadleafJobsEventDetail.java
cja769: That's the reason there's different constructors but I'll make the friendly name optional as well

common/src/main/java/org/broadleafcommerce/common/event/BroadleafJobsEventDetail.java
phillipuniverse: if value.lenght() > 255 then longValue, right?

cja769: I'm dumb. Didn't realize longValue meant a String greater than 255. I thought it meant long as in a number type.

cja769: 👍

phillipuniverse:
@cja769 I _think_ that this actually does not work if you have an extension of Sku but I am not 100% sure. It would be for the same reasons as BroadleafCommerce/QA#404, we need to do a final test with an extension of Sku in the following scenarios:

1. Unique `@Cache` annotation (2 invalidation events should get sent for each type and their respective regions):

    ```java
    @Entity
    @Table(name = "CUSTOM_SKU")
    @Cache("newCache")
    public class CustomSku extends SkuImpl
    ```

2. Extension without an `@Cache` annotation (1 event targeting the super class?)

    ```java
    @Entity
    @Table(name = "CUSTOM_SKU")
    public class CustomSku extends SkuImpl
    ```

PR #1876, opened 2018-05-04
Validation not happening on dynamic forms (pages and structured content)

marieStandeven:
https://github.com/BroadleafCommerce/QA/issues/3321

I think the other changes went through, this is just the commit for the pull request to link it to the issue.


phillipuniverse:
2 parts to this:
1. `EntityValidatorService` is inaccurate in its documentation to what actually happens in the implementation. The interface and docs said that null was a valid case, but in fact it caused all validators to be skipped (we were passing null for the `instance` to the validator when validating our dynamic fields)
2. Even if there were validation issues, when the AJAX request came back with the erroneous page fields there was an issue with the jquery selector since the field contains `|`

These have now both been updated.

phillipuniverse:
@marieStandeven just a heads up, allowing null in the `EntityValidatorServiceImpl` seemed like a more dangerous change. I found multiple examples of our `PropertyValidator`s not checking for null values, and it seemed like it would be relatively easy to hit an NPE. The changes that I made instead I believe are a bit safer and less prone to side-effects.

PR #1884, opened 2018-05-11
On Offer form reset the offer template responds correctly to Discount Price List functionality

danielcolgrove:
Fixes https://github.com/BroadleafCommerce/QA/issues/3307


danielcolgrove:
This change is not needed.

PR #1886, opened 2018-05-11
Fix UseShippingAddress for mixed Shippable/Nonshippable orders

michaelkloc:
>FirstShippableFulfillmentGroup instead of fulfillmentGroup[0]


michaelkloc:
https://github.com/BroadleafCommerce/ReferenceSite/pull/111

PR #1887, opened 2018-05-14
Rebaseline across library upgrades - Spring 5, Spring Boot 2, Hibernate 5, JDK 8

phillipuniverse:
First major feature within Broadleaf 6.0 as a baselining effort across our core libraries. This generally targets the Spring Boot 2.0 dependency line which includes:

- Hibernate 5.2.17
- Spring 5.0.6
- Spring Security 5.0.5
- Spring Boot 2.0.2
- Jackson 2.9.5

In regards to Java versions, this rebaselines Broadleaf on JDK 8 and 6.0 will be the first release compatible with JDK 9.

Other changes in the Broadleaf 6.0 line will likely increment these as new versions become available.


phillipuniverse:
This change currently does not build on the build server because of failing integration tests. However in order to test out other dependency changes targeting Broadleaf at the compile level and avoid further drift/risk to the 6.0 line I am merging this in to simplify what we need to target across other dependencies.

PR #1889, opened 2018-05-14
Add Support for the Java bin RequestWriter during Solr indexing

cja769:
This fixes an issue when indexing Money objects when the HttpSolrClient's RequestWriter is the BinaryRequestWriter. Without this, fields are indexed as "price_d : 'org.broadleafcommerce.common.money.Money:3.0" instead of as a double. When using solrj 5.x this isn't typically an issue but there are situations with replication where this can occur. Furthermore this is more fool-proof and is required when using solrj 7


phillipuniverse:
Just adding a bit more information here.

Solr 5.x uses the `org.apache.solr.client.solrj.request.RequestWriter` in the `HttpSolrClient, which converts all requests/responses to XML. To achieve this, it runs `toString()` on all non-primitives to serialize to XML. This is used for interacting with standalone Solr. When you move to the `CloudSolrClient` it, by default, uses the `org.apache.solr.client.solrj.impl.BinaryRequestWriter` which attempts to directly serialize the Java objects, and barfs on the complex types.

Starting with Solr 6, they have moved the default of _all_ `SolrClient` implementations to the `BinaryRequestWriter`, mainly because it's faster. While you could definitely customize these clients and use the XML-based `RequestWriter` instead, utilizing the `toString()` method for the `BinaryRequestWriter` works just as well and reduces further issues if someone decides to use the binary format and it doesn't work.

PR #1890, opened 2018-05-14
Upgrade Solr client to 7.3.1 and remove remaining support for Sku indexing via solr.index.use.sku

cja769:
This change accomplishes 2 main things
- Upgrade Solr version to 7.3.0
- Remove `solr.index.use.sku`
    - This was removed since our sku browsing support is not what it should be so we're simply going to remove what support we have and start over

#### Things changed for compilation based on API changes from Solr 5.3 to 7.3
  - Creating a collection
    - Was: `new CollectionAdminRequest.Create().setCollectionName(collectionName).setNumShards(numShards).setConfigName(solrConfigName)`
    - Now: `CollectionAdminRequest.create(collectionName, solrConfigName, numShards, numReplicas)`
  - Creating a collection alias
    - Was: `new CollectionAdminRequest.CreateAlias().setAliasName(reindex.getDefaultCollection()).setAliasedCollections(collectionName)`
    - Now: `CollectionAdminRequest.createAlias(aliasName, collectionName)`
- CloudSolrClient contructors changed
    - Now it takes a CloudSolrClient#Builder object

### DIFF
--- a/core/broadleaf-framework-web/src/main/java/org/broadleafcommerce/core/web/controller/cart/BroadleafCartController.java
+++ b/core/broadleaf-framework-web/src/main/java/org/broadleafcommerce/core/web/controller/cart/BroadleafCartController.java
@@ -68,9 +68,6 @@ public class BroadleafCartController extends AbstractCartController {
 
     protected static String ALL_PRODUCTS_ATTRIBUTE_NAME = "blcAllDisplayedProducts";
 
-    @Value("${solr.index.use.sku}")
-    protected boolean useSku;
-
     @Value("${automatically.add.complete.items}")
     protected boolean automaticallyAddCompleteItems;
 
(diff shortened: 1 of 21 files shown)


core/broadleaf-framework-web/src/main/java/org/broadleafcommerce/core/web/controller/cart/BroadleafCartController.java
```diff
@@ -302,11 +299,7 @@ public String updateQuantity(HttpServletRequest request, HttpServletResponse res
         
         if (isAjaxRequest(request)) {
             Map<String, Object> extraData = new HashMap<>();
-            if(useSku) {
-                extraData.put("skuId", itemRequest.getSkuId());
-            } else {
-                extraData.put("productId", itemRequest.getProductId());
-            }
+            extraData.put("productId", itemRequest.getProductId());
```
phillipuniverse: This is not quite right, you should just always set the `skuId` based on the `itemRequest` anyway.

phillipuniverse (changes requested):

PR #1891, opened 2018-05-16
Fixes the ability to sort by price after adding a price filter.

marieStandeven:
https://github.com/BroadleafCommerce/QA/issues/3331

Fixes the ability to sort by price after adding a price filter.


danielcolgrove:
This seems to be fixed.  It's working on Demo Site.

PR #1898, opened 2018-05-29
Internalization enhancements

oleksiimiroshnyk:
Internalization enhancements
show/hide translated


phillipuniverse (changes requested):
Technically, in 5.2 we still support Thymeleaf 2 in the admin. Because of this you should add an `th:inline="text"` attribute to each of these elements which will just be ignored in Thymeleaf 3.

Broadleaf 6.0+ we drop support for Thymeleaf 2 altogether.

phillipuniverse:
Also, the build failure will resolve itself it failed because of a problem with the build server.

PR #1900, opened 2018-05-29
Bundle Dependency Management Enhancements and Preloading

jmitash:
Enhancements to bundle dependency management
-------------------
Referring to the `bundle-dependency-event` attribute on `<blc:bundle>` tags which will add the script to the DOM once the specified event fires:
- The dependency handling JavaScript no longer requires jQuery
- Now works with unbundled files and ensures they are executed in order (except when the files are async/deferred)

New feature: bundle complete events and dependency chaining
-------------------
Bundles can now be chained together to run in order (and asynchronously) so that their dependencies will be satisfied when they run.

A new attribute for `<blc:bundle>` tags, `bundle-completed-event`, has been introduced. This allows you to specify an event to fire when a bundle is complete. It is also compatible with `bundle-dependency-event` so that a bundle can have a dependency and fire an event when complete. The only exception is when bundling is disabled, the scripts are set to be added asynchronously, and the bundle has a `bundle-dependency-event` in which case `bundle-completed-event` is unsupported since the JavaScript can't tell when the bundle is complete.

This new attribute along with `bundle-dependecy-event` allows chaining of bundles as dependencies. For example, if `3.js` must run after `2.js` which must run after `1.js`, this can be expressed as

```
<blc:bundle name="1.js" bundle-completed-event="1Finished" async ... />
<blc:bundle name="2.js" bundle-dependency-event="1Finished" bundle-completed-event="2Finished" async ... />
<blc:bundle name="3.js" bundle-dependency-event="2Finished" async ... />
```

in which case, the bundles will run in the order of 1, 2, 3 despite being asynchronous.

Introduction of \<blc:bundlepreload\>
-------------------
The above example can introduce a problem with page speed as wherever `bundle-dependency-event` is used, the bundle is not in the DOM until the dependency is met, and only then will it be added to the DOM by the dependency-handling JavaScript. This can make the site slower as the browser won't start downloading a JavaScript file until it's added to the DOM. To counter this problem, preloading can be used to tell the browser to download a file before it is needed:
```
<link rel="preload" href="/js/dependent-bundle.js" as="script" />
```
which is the functionality that the `<blc:bundlepreload>` provides by generating these link tags for the bundle.

However, using this with bundles can be messy since the list of files would need to be duplicated for the `<blc:bundlepreload>` and `<blc:bundle>` tags. To solve this problem, the bundle (or list of files if bundling is disabled) is saved on the request whenever a `<blc:bundle>` or `<blc:bundlepreload>` tag is parsed. Any following `<blc:bundlepreload>` and `<blc:bundle>` that need the same list of files can leave off the `files` using only the `name` attribute to refer to the same bundle.

So instead of having this:
```
<blc:bundlepreload
    name="dependent-bundle.js"
    mapping-prefix="/js/"
    files="1.js,
    2.js,
    3.js,
    4.js,
    5.js,
    ...,
    100.js" />

<blc:bundle name="dependent-bundle.js"
    mapping-prefix="/js/"
    async="true"
    defer="true"
    bundle-dependency-event="dependencyEvent"
    files="1.js,
    2.js,
    3.js,
    4.js,
    5.js,
    ...,
    100.js" />
```

The attributes used to generate the bundle only need to appear on the first tag for each bundle:
```
<blc:bundlepreload
    name="dependent-bundle.js"
    mapping-prefix="/js/"
    files="1.js,
    2.js,
    3.js,
    4.js,
    5.js,
    ...,
    100.js" />

<blc:bundle name="dependent-bundle.js" async="true" defer="true" bundle-dependency-event="dependencyEvent" />
```

Note that when bundling is enabled, `bundle-completed-event`s **must** be on both the `<blc:bundlepreload>` and `<blc:bundle>` tags since it affects how the bundle is generated.

Non-render-blocking CSS Support
-------------------
CSS bundles can now be handled in a non-render-blocking fashion by setting `defer="true"`. The CSS will download asynchronously with the help of `<link rel="preload">` tags before being transformed to normal `rel="stylesheet"` when they've downloaded. A `<noscript>` fallback is included to render the CSS normally.

BroadleafCommerce/QA#3344

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceBundlingService.java
+++ b/common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceBundlingService.java
@@ -10,7 +10,7 @@
  * the Broadleaf End User License Agreement (EULA), Version 1.1
  * (the "Commercial License" located at http://license.broadleafcommerce.org/commercial_license-1.1.txt)
  * shall apply.
- *
+ * 
  * Alternatively, the Commercial License may be replaced with a mutually agreed upon license (the "Custom License")
  * between you and Broadleaf Commerce. You may not use this file except in compliance with the applicable license.
  * #L%
--- a/common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceBundlingService.java
+++ b/common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceBundlingService.java
@@ -10,7 +10,7 @@
  * the Broadleaf End User License Agreement (EULA), Version 1.1
  * (the "Commercial License" located at http://license.broadleafcommerce.org/commercial_license-1.1.txt)
  * shall apply.
- * 
+ *
  * Alternatively, the Commercial License may be replaced with a mutually agreed upon license (the "Custom License")
  * between you and Broadleaf Commerce. You may not use this file except in compliance with the applicable license.
  * #L%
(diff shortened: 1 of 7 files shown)


jmitash:
[(2.5) jQuery occasionally doesn't load before being called due to race condition between JS dependencies [2]](https://trello.com/c/m8TFCyNx/12-025-jquery-occasionally-doesnt-load-before-being-called-due-to-race-condition-between-js-dependencies-2)

ckittrell (approved):

phillipuniverse:
@knoxcorner @ckittrell I detected a regression here in the following case, consider this in `head.html`:

```html
<blc:bundle name="bundle.css"
                        mapping-prefix="/css/"
                        files="custom-css-1.css"/>

<!-- ... later on in the same file, or in a fragment or whatever ... -->

<blc:bundle name="bundle.css"
                        mapping-prefix="/css/"
                        files="custom-css-2.css"/>
```

Before this change (5.2.4-GA), you would get a `bundle.css` file (note that both of the bundle names are the same) with both `custom-css-1.css` _and_ `custom-css-2.css`. With these changes, you only get `custom-css-1.css`. I specifically noticed this when it was unbundled; `custom-css-2.css` never even appears in the final rendered HTML.

I am still testing through this a bit to see if I can figure out what's going on in general, but can you guys help validate this regression tomorrow?

jmitash:
@phillipuniverse That was an oversight on my part. I wasn't aware it was possible to have multiple bundles with the same name, so when it gets to the second bundle it thinks it's referring to the first one. I'll have to experiment with 5.2.4 to see how it handles that situation and get it to check the files regardless of the bundle name.

phillipuniverse:
Follow-up to this is at #1935.

PR #1911, opened 2018-06-06
Move Tika dependency to CMS only, and remove the MimeUtils dependency

epenning:
Originated at https://github.com/BroadleafCommerce/BroadleafCommerce/pull/1854#issuecomment-384110546

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1183,25 +1183,6 @@
                 <version>1.8.13</version>
             </dependency>
             <dependency>
-                <groupId>eu.medsea.mimeutil</groupId>
-                <artifactId>mime-util</artifactId>
-                <version>2.1.3</version>
-                <exclusions>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -1182,30 +1182,6 @@
...
-                        <groupId>log4j</groupId>
-                        <artifactId>log4j</artifactId>
-                    </exclusion>
-                </exclusions>
-            </dependency>
-            <dependency>
-                <groupId>org.apache.tika</groupId>
-                <artifactId>tika-core</artifactId>
-                <version>1.17</version>
-            </dependency>
```
cja769: We want to leave this here so that the version is declared globally. We use the parent pom this way to a) have a singular place for versioning and b) so that if anyone wants a dependency they automatically get the version we support

PR #1912, opened 2018-06-07
Introduce Minimum Order Quantity Validation for Skus

ckittrell:
For the `5.2.5` version of this feature, the `enable.sku.minOrderQuantity.field` must be set to `true` to opt in since there are domain updates for the `BLC_SKU` table.

This feature allows you to specify the minimum allowable quantity for a Sku to be in your cart. For instance, if you set the minimum quantity to 10, then customers must add at least 10 of these items to their cart and cannot update the quantity of this item to be less than 10.


Jerry77oz (approved):

ckittrell:
[(1) Introduce "Min Quantity Per Order" Sku concept [0.75]](https://trello.com/c/gcGI80N3/45-1-introduce-min-quantity-per-order-sku-concept-075)

PR #1916, opened 2018-06-11
QA-3354 Product groups/add-ons not working correctly

oleksiimiroshnyk:
BroadleafCommerce/QA#3354
Apply typed enity filter for product more correctly


ccs19:
Looks like add-ons are showing up correctly now. However, I noticed that they are not showing up under "All Products". The count accurately reflects Base + Add On products, but Add Ons are not loading and showing as `(No records found)`. See attached image.
image

ccs19:
There is also another bug regarding Product Groups. When using rule based groups, e.g. Products in a category, add-ons don't seem to be included in the results. Base products work fine with the rule based product groups.

oleksiimiroshnyk:
@ccs19 can't reproduce issue with All Products.
allproducts

Not really familiar with Product Groups, can please give me an overview and steps to reproduce?

ccs19:
Sure.

Go to Catalog -> Product Groups -> Add a new group -> Select Rule Based -> Product Category (is in) A category

Then attach that product group to a product as an add-on.

PR #1935, opened 2018-06-20
Fix issue where multiple bundles with same name get the same file lists

jmitash:
BroadleafCommerce/QA#3404

The changes to resource bundle processing that allow for some of the tag attributes to be left off when referring to the same bundle were causing bundles with the same name to always have the file list of the first bundle with that name on the template. The implicit tag attributes will now check the mapping prefix and file list if provided.


phillipuniverse:
Confirmed fixed, thanks!

phillipuniverse:
Note: this is a follow-up to #1900.

PR #1936, opened 2018-06-20
Reenable save button after to-one selection

jmitash:
BroadleafCommerce/QA#3384

The save button was not being reenabled after changing a field that requires a to-one lookup after a validation failure on DemoSite.


cja769:
This was validated by @Hashmi42 that this was a pre-existing bug in 5.2 so I rebased the branch on top of 5.2. He also made sure the fix works in 5.2 and 6.0 as well as it didn't break ReferenceSite since the bug this fixes only exists in DemoSite

PR #1937, opened 2018-06-20
Added minimum quantity exception and removed error logging

marieStandeven:
https://github.com/BroadleafCommerce/ReferenceSite/pull/136


ckittrell (changes requested):
@marieStandeven When the exception is thrown, it needs to be given the Product's id.

ckittrell (approved):

PR #1943, opened 2018-06-22
Added functionality to combine the indexes defined on the weave class with the target classes @Table annotation

cja769:
This was to alleviate a bug where having the Hibernate `@Index` on a column in an embeddable class resulted in a single index with all of the columns added by the weave instead of just adding an index for the specified columns in Hibernate 5+. The implementation of the merge was inspired by how we merge the `@EntityListeners` with some tweaks since the indexes are an array of annotations instead of an array of classes. With the change we should no longer add Hibernate `@Index` annotations to columns on embedded classes and instead add a JPA `@Table` annotation on the weave class. The `@Table` annotation won't cause any side-affects since the weave classes aren't hooked up to a persistence unit. However, if in the future we choose to register domain objects with class scanning then we'll have to simply create a new annotation for storing the indexes and update the `DirectCopyClassTransformer`.

This solves https://github.com/BroadleafCommerce/QA/issues/3408

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/extensibility/jpa/copy/DirectCopyClassTransformer.java
+++ b/common/src/main/java/org/broadleafcommerce/common/extensibility/jpa/copy/DirectCopyClassTransformer.java
@@ -41,6 +41,8 @@ import java.util.Set;
 
 import javax.annotation.Resource;
 import javax.persistence.EntityListeners;
+import javax.persistence.Index;
+import javax.persistence.Table;
 
 import javassist.ClassPool;
 import javassist.CtClass;
@@ -200,6 +202,7 @@ public class DirectCopyClassTransformer extends AbstractClassTransformer impleme
                     ClassFile templateFile = template.getClassFile();
                     ConstPool constantPool = classFile.getConstPool();
                     buildClassLevelAnnotations(classFile, templateFile, constantPool);
+                    transformIndexAnnotations(classFile, templateFile, constantPool);
 
                     // Copy over all declared fields from the template class
                     // Note that we do not copy over fields with the @NonCopiedField annotation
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/broadleafcommerce/common/extensibility/jpa/copy/DirectCopyClassTransformer.java
```diff
@@ -488,6 +491,114 @@ protected XFormParams reviewConditionalDirectCopyTransforms(String convertedClas
...
+                    for (Annotation annotation : items) {
+                        String typeName = annotation.getTypeName();
+                        if (typeName.equals(Table.class.getName())) {
+                            logger.debug("Stripping out previous Table annotation at the class level - will merge into new Table");
+                            existingTable = annotation;
+                            continue;
+                        }
+                        annotationsAttribute.addAnnotation(annotation);
+                    }
+                    itr.remove();
```
phillipuniverse, line 531: This appears to remove _all_ annotations from the `@Entity`. Seems like you should just be removing the `@Table`, right? What exactly are you removing here? Little difficult to follow (for good reason, this is difficult to work with).

cja769: @phillipuniverse An `AnnotationAttribute` is a basket that holds all of the annotations. That `itr.remove()` doesn't remove the `@Table` or `@Entity` annotation, it removes all class level annotations. However the `annotationsAttribute` list holds all of those annotation to be re-added later. This piece of code was copied from how we merge `@EntityListeners`

phillipuniverse: Got it. So it removes everything, but also keeps track of everything that it removed, except for the `@Table` annotation. That stays removed, but you hold onto it with the `existingTable` variable. Then you modify it below (or more technically, create a new one with your changes), and then add it to the `annotationsAttribute`, which gets added back as class-level annotations.

cja769: You got it dude!

PR #1944, opened 2018-06-25
Added null check for order when DatabaseOrderLockManager is configured to throw exception instead of queueing the lock attempt

danielcolgrove:
This will get called with order.lock.errorInsteadOfQueue=true.  Fixes https://github.com/BroadleafCommerce/QA/issues/3411

### DIFF
--- a/core/broadleaf-framework-web/src/main/java/org/broadleafcommerce/core/web/order/DatabaseOrderLockManager.java
+++ b/core/broadleaf-framework-web/src/main/java/org/broadleafcommerce/core/web/order/DatabaseOrderLockManager.java
@@ -87,7 +87,7 @@ public class DatabaseOrderLockManager implements OrderLockManager {
 
     @Override
     public Object acquireLockIfAvailable(Order order) {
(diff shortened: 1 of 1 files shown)


core/broadleaf-framework-web/src/main/java/org/broadleafcommerce/core/web/order/DatabaseOrderLockManager.java
```diff
@@ -87,7 +87,7 @@ public Object acquireLock(Order order) {
 
     @Override
     public Object acquireLockIfAvailable(Order order) {
-        if (order instanceof NullOrderImpl) {
+        if (order == null || order instanceof NullOrderImpl) {
             if (LOG.isDebugEnabled()) {
                 LOG.debug("Attempted to grab a lock for a NullOrderImpl. Not blocking");
```
Jerry77oz: Should the debug message also be updated to say that the order could be null?

Jerry77oz (approved):

PR #1946, opened 2018-06-27
Need to be able to view added filters

hasheniuk:
https://trello.com/c/LSy6AqDv/17-3-need-to-be-able-to-view-added-filters
View applied filters above the main entity list grid

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/css/admin/listGrid.css
+++ b/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/css/admin/listGrid.css
@@ -52,6 +52,27 @@
     position: relative;
     top: 6px;
 }
+.listgrid-search .filter-info .button-group {
+    position: relative;
+    top: -28px;
+}
+.listgrid-search .filter-pillow-wrapper {
+    display: none;
+    width: calc(100% - 111px);
+    height: 46px;
+    overflow-y: hidden;
+    overflow-x: auto;
+}
+.listgrid-search .filter-pillow-wrapper .filter-pillow-container {
+    width: auto;
+    padding: 4px;
+}
+.listgrid-search .filter-pillow-wrapper .filter-pillow-container .filter-pillow {
+    border-radius: 5px;
+}
+.listgrid-search .filter-pillow-wrapper .filter-pillow-container .filter-pillow:not(:last-child) {
+    margin-right: 5px;
+}
 .listgrid-search-actions {
     display: inline-block;
 }
@@ -252,6 +273,31 @@ div.alert-box.list-grid-alert.save-alert a.close {
 }
 .listgrid-toolbar .filter-info .clear-filters i {
     padding-top: 0;
+}
+.listgrid-toolbar .filter-info .button-group {
+    vertical-align: top;
+}
+.listgrid-toolbar .filter-pillow-wrapper {
+    display: none;
+    vertical-align: top;
+    max-width: 140px;
+    overflow-y: hidden;
+    overflow-x: auto;
+}
+@media (min-width: 1200px) {
+    .listgrid-toolbar .filter-pillow-wrapper {
+        max-width: 300px;
+    }
+}
+.listgrid-toolbar .filter-pillow-wrapper .filter-pillow-container {
+    padding: 2px;
+    overflow: auto;
+}
+.listgrid-toolbar .filter-pillow-wrapper .filter-pillow-container .filter-pillow {
+    border-radius: 5px;
+}
+.listgrid-toolbar .filter-pillow-wrapper .filter-pillow-container .filter-pillow:not(:last-child) {
+    margin-right: 5px;
 }
 /****************************************
  ADMIN LISTGRID/TABLES STYLES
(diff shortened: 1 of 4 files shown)


admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/filterbuilder.js
```diff
@@ -907,6 +907,62 @@
...
+                switch(rule.operator) {
+                    case "IS_NULL":
+                        break;
+                    case "BETWEEN":
+                        var arr = JSON.parse(rule.value);
+                        value = "<b>" + arr[0] + " AND " + arr[1] + "</b>";
+                        break;
+                    case "COLLECTION_IN":
+                    case "COLLECTION_NOT_IN":
+                        var arr = JSON.parse(rule.value);
```
nathandmoore: @hasheniuk, to get the name for the collection-in/not-in, 
1. Grab the entity context path from `$('#listGrid-main-header).data('path')`
2. Grab the field name from the rule (`rule.field`)
3. Create a new endpoint in `AdminBasicEntityController` like `/selectize/name`
4. Send the field name and value (e.g., defaultCategory:2001) to the new endpoint
5. Add a method in `AdminEntityService` that can take the sectionKey (/admin/product), the field name (defaultCategory), and the ID to return the object
6. then take the name from it 
7. and return it

I think you may need to add `blDynamicEntityDao` to the AdminEntityService and add a method to it as well that works like `DynamicEntityDao#readOtherEntitiesWithPropertyValue()` but returns the associated entity/value for the field in question instead of a list of IDs

admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/css/admin/listGrid.css
```diff
@@ -33,6 +33,25 @@
     height: 45px;
     white-space: nowrap;
 }
+.listgrid-search .applied-filters-container {
```
nathandmoore: Let's make the added filters look akin to bootstrap nav pills (https://getbootstrap.com/docs/4.0/components/navs/#pills). Note that we don't have bootstrap 4 in open-admin, so you'll need to do it custom. You can probably borrow some of the styling from `selectize-admin-customization.css`:

```css
.selectize-control.multi .selectize-input [data-value] {
    background-color: #8EAA2B;
    border:none;
    margin: 5px 0 5px 5px;
}
```

nathandmoore: Also, we need to check what these added filters look like when using a listGrid for a collection and in a modal. See Product custom attributes (bottom of first tab) and ProductOption lookup on Product's "Options" tab

nathandmoore (changes requested):

admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/filterbuilder.js
nathandmoore: @jfleschler Does that sound about right?

nathandmoore: We will only use this method for `COLLECTION_IN` or `COLLECTION_NOT_IN`, so when we build the criteria query we can do the following:

1. Get the classname for the sectionKey
2. Get the Impl class for the classname (see `DynamicEntityDao#getImplClass()`
3. Use reflection to get the Impl class (ProductImpl) of the field (defaultCategory -> CategoryImpl)
4. Setup a CriteriaQuery = builder.createQuery(CategoryImpl)
5. Root = criteria.from(ProductImpl)
6. Restriction = builder.in(root.get(fieldName).get("id"), <list of ids from the filter>)
etc.

We'll always be appending the `.get("id")` part

admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/css/admin/listGrid.css
nathandmoore: If we have something like `COLLECTION_IN` rule, then the pill could look like `Parent Category in: Home, Hot Sauces, Merchandise`

admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/filterbuilder.js
hasheniuk: @nathandmoore 
In `AdminBasicEntityController` I found similar method for getting `COLLECTION_IN` names `viewEntityListSelectize`.

admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/css/admin/listGrid.css
hasheniuk: @nathandmoore 
I applied styles similar to selectize input and also I attached new screenshots to Trello issue.

nathandmoore (approved):

PR #1948, opened 2018-06-27
Cleaned up AdminUserDetails, Service, and AdminProvisioningService

nathandmoore:
Affects BroadleafCommerce/QA#3113


phillipuniverse:
This seems a bit strange to muddy the waters here, where there is no knowledge of "approvals" at all, instead of just in BroadleafCommerce/Enterprise#982. Why does this code need to be here in core and not only in Enterprise?

nathandmoore:
I saw that Sandbox and Multitenant copy domains are already in common for example and just figured that if it was like that already, I could avoid adding multiple enterprise extensions. But I'll move it since that's not really a strong argument.

jfleschler (approved):

PR #1956, opened 2018-07-06
Performance improvements for building metadata

oleksiimiroshnyk:
performance improvements for building metadata after profiling
Originally was asked by emeratech that return request screen is slow.
This is because new entity was introduced and it has field orderItem, fulfillmentOrderItem etc, and some of the fields that should be visible in grid are deep inside orderItem.xxx.yyy.zzz.fieldName
So to build metadata for those properties metadata buiilder does a pretty deep recursion and it takes some time to build metadata for the introduced class.
Profiling with invocation count and time spent identified couple of hot spots that I was trying to address in that PR. in general reflection is not fast thing, so some caching can be handy.

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/dao/DynamicEntityDaoImpl.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/dao/DynamicEntityDaoImpl.java
@@ -153,6 +153,9 @@ public class DynamicEntityDaoImpl implements DynamicEntityDao, ApplicationContex
 
     protected ApplicationContext applicationContext;
 
+
+    protected FieldManager fieldManager = new FieldManager(entityConfiguration, null);
+
     @Override
     public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
         this.applicationContext = applicationContext;
@@ -1554,6 +1557,7 @@ public class DynamicEntityDaoImpl implements DynamicEntityDao, ApplicationContex
     @Override
     public void setStandardEntityManager(EntityManager entityManager) {
         this.standardEntityManager = entityManager;
+        fieldManager = new FieldManager(entityConfiguration, entityManager);
     }
 
     @Override
@@ -1568,7 +1572,7 @@ public class DynamicEntityDaoImpl implements DynamicEntityDao, ApplicationContex
 
     @Override
     public FieldManager getFieldManager() {
-        return new FieldManager(entityConfiguration, getStandardEntityManager());
+        return fieldManager;
     }
 
     @Override
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/dao/DynamicEntityDaoImpl.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/dao/DynamicEntityDaoImpl.java
@@ -154,7 +154,7 @@ public class DynamicEntityDaoImpl implements DynamicEntityDao, ApplicationContex
     protected ApplicationContext applicationContext;
 
 
-    protected FieldManager fieldManager = new FieldManager(entityConfiguration, null);
+    protected FieldManager fieldManager;
 
     @Override
     public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
@@ -1572,6 +1572,11 @@ public class DynamicEntityDaoImpl implements DynamicEntityDao, ApplicationContex
 
     @Override
     public FieldManager getFieldManager() {
+        if (fieldManager == null) {
+            //keep in mind that getStandardEntityManager() can return null, this is in general OK,
+            // we re-init fieldManager in setStandardEntityManager method
+            fieldManager = new FieldManager(entityConfiguration, getStandardEntityManager());
+        }
         return fieldManager;
     }
 
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/FieldManager.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/FieldManager.java
@@ -62,7 +62,7 @@ public class FieldManager {
     protected EntityConfiguration entityConfiguration;
     protected EntityManager entityManager;
     protected List<SortableValue> middleFields = new ArrayList<SortableValue>(5);
-    protected Set<Class> classes = new HashSet<>();
+    protected Set<Class> managedEntityClasses = new HashSet<>();
     protected Map<Class, Map<String, Field>> classFields = new HashMap<>();
 
     public FieldManager(EntityConfiguration entityConfiguration, EntityManager entityManager) {
@@ -71,7 +71,7 @@ public class FieldManager {
         if (entityManager != null) {
             Set<EntityType<?>> managedEntities = entityManager.getMetamodel().getEntities();
             for (EntityType managedEntity : managedEntities) {
-                classes.add(managedEntity.getJavaType());
+                managedEntityClasses.add(managedEntity.getJavaType());
             }
         }
     }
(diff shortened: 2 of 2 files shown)


admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/FieldManager.java
```diff
@@ -48,37 +49,57 @@
...
-        SessionFactory sessionFactory = dynamicEntityDao.getDynamicDaoHelper().
-                getSessionFactory((HibernateEntityManager) dynamicEntityDao.getStandardEntityManager());
-        BLCFieldUtils fieldUtils = new BLCFieldUtils(sessionFactory, true, dynamicEntityDao.useCache(),
-                dynamicEntityDao.getEjb3ConfigurationDao(), entityConfiguration,
-                dynamicEntityDao.getDynamicDaoHelper());
-        return fieldUtils.getField(clazz, fieldName);
+        Map<String, Field> stringFieldMap = classFields.get(clazz);
+        if(stringFieldMap==null){
+            stringFieldMap = new HashMap<>();
+            classFields.put(clazz, stringFieldMap);
```
phillipuniverse: Do we need to be worried about thread safety here?

Also, I looked over in `BLCFieldUtils` and that already has some caching there for `Field` instances (in `getSingleField()`). Maybe the part that it's missing is the class -> `Field` mapping? Or that the caching is not around `getField()`?

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/dao/DynamicEntityDaoImpl.java
```diff
@@ -153,6 +153,9 @@
 
     protected ApplicationContext applicationContext;
 
+
+    protected FieldManager fieldManager = new FieldManager(entityConfiguration, null);
```
phillipuniverse: Why not start this as `null` and make it behave the same way as the `standardEntityManager`? It looks like you are intending them to have the same semantics and they are essentially tied together.

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/FieldManager.java
```diff
@@ -48,37 +49,57 @@
...
     protected EntityManager entityManager;
     protected List<SortableValue> middleFields = new ArrayList<SortableValue>(5);
+    protected Set<Class> classes = new HashSet<>();
+    protected Map<Class, Map<String, Field>> classFields = new HashMap<>();
 
     public FieldManager(EntityConfiguration entityConfiguration, EntityManager entityManager) {
         this.entityConfiguration = entityConfiguration;
         this.entityManager = entityManager;
+        if (entityManager != null) {
+            Set<EntityType<?>> managedEntities = entityManager.getMetamodel().getEntities();
```
phillipuniverse, line 72: 👍 solid improvement for sure, no need to re-figure this out every time!

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/FieldManager.java
```diff
@@ -48,37 +49,57 @@
...
-    
+
     private static final Log LOG = LogFactory.getLog(FieldManager.class);
 
     public static final String MAPFIELDSEPARATOR = "---";
 
     protected EntityConfiguration entityConfiguration;
     protected EntityManager entityManager;
     protected List<SortableValue> middleFields = new ArrayList<SortableValue>(5);
+    protected Set<Class> classes = new HashSet<>();
```
phillipuniverse: Maybe rename this variable to `managedEntityClasses` to provide better traceability into what this is doing?

phillipuniverse (changes requested):

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/FieldManager.java
oleksiimiroshnyk: I was thinking about concurrency here and what I found is that the entry point was a prototype bean, so for each reqeust it should start new instance of all that stuff and no instances are shared between threads. But that's one of things I was hoping you/Jeff can point out if I miss something in the flows where this piece of code is used.

About BLCFieldUtils, hmm, I missed that it already has a cache but profiler didn't reveal that this cache is really used. I think because for each call of a org.broadleafcommerce.openadmin.server.service.persistence.module.FieldManager#getField new instance of `BLCFieldUtils` is created.
```java
            SessionFactory sessionFactory = dynamicEntityDao.getDynamicDaoHelper().
                    getSessionFactory((HibernateEntityManager) dynamicEntityDao.getStandardEntityManager());
            BLCFieldUtils fieldUtils = new BLCFieldUtils(sessionFactory, true, dynamicEntityDao.useCache(),
                    dynamicEntityDao.getEjb3ConfigurationDao(), entityConfiguration,
                    dynamicEntityDao.getDynamicDaoHelper());
            Field fld = fieldUtils.getField(clazz, fieldName);
```
Probably my cache can be removed and instance of `BLCFieldUtils` is created per `FieldManager` instance, if at `FieldManager` construction time it has all needed properties, which might be a little tricky.

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/dao/DynamicEntityDaoImpl.java
oleksiimiroshnyk: Hmm, don't rembmer, I think main reason behind that if in some new place a `DynamicEntityDaoImpl` will be created it is not really necessary to call the `setStandardEntityManager` method, and potentially we might have NPE... I can probably move this instatiation to `getFieldManager()` method but not sure what I don't like more setters or getters that do some extra work.
Maybe DynamicEntityDao should have a constructor for that field, but I'm not really sure about all the places where and how it can/will be used, so tried to do non-destructive change
But you are right, the main reason is to tight `standardEntityManager` and `fieldManager` , so let me know if I should default it to null and maybe move "check for null and instance creation as a  protection" to getter. Just in case

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/FieldManager.java
oleksiimiroshnyk: checked for null and create one instance of BLCFieldUtils per FieldManager instance

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/dao/DynamicEntityDaoImpl.java
phillipuniverse: Yeah I think default to null and move the null-checking protection in the getter makes the most sense.

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/FieldManager.java
phillipuniverse: > I was thinking about concurrency here and what I found is that the entry point was a prototype bean, so for each reqeust it should start new instance of all that stuff and no instances are shared between threads.

That's true, should be ok.

> About BLCFieldUtils, hmm, I missed that it already has a cache but profiler didn't reveal that this cache is really used. I think because for each call of a org.broadleafcommerce.openadmin.server.service.persistence.module.FieldManager#getField new instance of BLCFieldUtils is created.

Yes, but the `FIELD_CACHE` is static (and usage is in a `synchronized` block) which means that it's shared across all instances. It seems like it is more likely that in the particular flow you were looking at is not actually exercising anything in the `FIELD_CACHE` for whatever reason.

Regardless, if your change here to have a single instance of the `FieldManager` still resolves the perf issues you were seeing then I think you're ok.

PR #1963, opened 2018-07-24
Clear the FieldManager's middle fields on a call to `getFieldManager()`

jfleschler:


Jerry77oz (approved):

nathandmoore:
@jfleschler What kind of bug did you see because of this?

jfleschler:
@nathandmoore I was recieving an error trying to save ContentItem's I believe.  Something about `Column NAME cannot be null`.  I traced it down to recent changes with the `getFieldManager()` where during the `FieldManager#persistMiddleEntities()` it wasn't clearing out the list of middle entities between different entity types.

nathandmoore:
Caused by https://github.com/BroadleafCommerce/BroadleafCommerce/pull/1956

PR #1991, opened 2018-10-09
Update minification to skip files that have min in their file extension

StanislavFedorov:
Google closure compiler can have issues parsing and minfying files that are already minified so it's better to just skip those files.

BroadleafCommerce/QA#3453

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceMinificationServiceImpl.java
+++ b/common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceMinificationServiceImpl.java
@@ -46,6 +46,7 @@ public class ResourceMinificationServiceImpl implements ResourceMinificationServ
     protected static final String COULD_NOT_MINIFY_RESOURCES_RETURNED_UNMINIFIED_BYTES = "Could not minify resources, returned unminified bytes";
     public static String CSS_TYPE = "css";
     public static String JS_TYPE = "js";
+    public static String JS_MIN = ".min.js";
 
     @Autowired
     protected Environment environment;
@@ -94,6 +95,19 @@ public class ResourceMinificationServiceImpl implements ResourceMinificationServ
             LOG.warn("Attempted to modify resource without a filename, returning non-minified resource");
             return originalResource;
         }
+
+        Integer filenameLength = originalResource.getFilename().length();
+
+        if (filenameLength >= JS_MIN.length()) {
+
+            String lastSevenCharacters = originalResource.getFilename().substring(filenameLength-JS_MIN.length());
+            if (lastSevenCharacters.equals(JS_MIN)) {
+                LOG.info("Minification already have done for this resource: " + originalResource.getFilename());
+                return originalResource;
+            }
+
+        }
+
         return minify(originalResource, originalResource.getFilename());
     }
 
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceMinificationServiceImpl.java
```diff
@@ -46,6 +46,7 @@
     protected static final String COULD_NOT_MINIFY_RESOURCES_RETURNED_UNMINIFIED_BYTES = "Could not minify resources, returned unminified bytes";
     public static String CSS_TYPE = "css";
     public static String JS_TYPE = "js";
+    public static String JS_MIN = ".min.js";
```
cja769, line 52: Should check .min.css as well since there's no point in minifying it as well even if YUICompressor doesn't have an issue with it

common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceMinificationServiceImpl.java
```diff
@@ -94,6 +95,19 @@ public Resource minify(Resource originalResource) {
...
+        Integer filenameLength = originalResource.getFilename().length();
+
+        if (filenameLength >= JS_MIN.length()) {
+
+            String lastSevenCharacters = originalResource.getFilename().substring(filenameLength-JS_MIN.length());
+            if (lastSevenCharacters.equals(JS_MIN)) {
+                LOG.info("Minification already have done for this resource: " + originalResource.getFilename());
+                return originalResource;
+            }
+
```
cja769: We don't need the check here because we're just delegating to the other `minify` method anyways. All this method does is make sure that the file has a name before delegating

common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceMinificationServiceImpl.java
```diff
@@ -110,6 +124,18 @@ public Resource minify(Resource originalResource, String filename) {
...
+        Integer filenameLength = originalResource.getFilename().length();
+
+        if (filenameLength >= JS_MIN.length()) {
+
+            String lastSevenCharacters = originalResource.getFilename().substring(filenameLength-JS_MIN.length());
+            if (lastSevenCharacters.equals(JS_MIN)) {
+                LOG.info("Minification already have done for this resource: " + filename);
+                return originalResource;
+            }
+
```
cja769: It's better to have this logic in its own method so that if a client wants to determine what is or what isn't a minified file they can do that easily. Additionally this can all be replaced by

```
if (StringUtils.endsWith(filename, JS_MIN)) {
    // Info log and return resource
}
```
Typically there's an apache commons StringUtils method for most String operations. This goes for Collections, Integers, Longs, Arrays, etc. It's a good habit to check if there's an apache commons method you can use instead so that we don't have to reinvent the wheel. If you have a feeling that there's probably a library that does something but you can't think of what the static class would be just ask in the Slack channel "developers" and someone will be able to help.

cja769 (commented):
I've already made these changes locally since they're pretty straightforward so I'll go ahead and make the changes and merge them

PR #1998, opened 2018-10-16
Fix NPE when viewing classes in the Approvals section that do not have an AdminSection attached to them

Jerry77oz:
BroadleafCommerce/QA#3486

- Add Null check before using the sections


Jerry77oz:
Closing this since we have a PR that was submitted here https://github.com/BroadleafCommerce/BroadleafCommerce/pull/2000

PR #2007, opened 2018-10-24
Fix template error when opening admin reset password link

jmitash:
The reset password link has no "messageCode" parameter which was previously causing an error when opening the link. This corrects the logic for when the messageCode should be used to display the "email sent" message.


sesker (approved):
Proper null checking.

PR #2012, opened 2018-11-01
Remove collection alias' from SolrQuery in favor of executing commands on a collection

cja769:
After upgrading Solr to version 7 in BLC 6 we found that search requests were failing if you were using POST. It was found that this was being caused because we were sending the collection alias that we wanted to operate on via a SolrQuery parameter which is no longer allowed. The proper way to specify the collection you want to operate on is in the actual call through SolrJ (i.e. `server.query(collection, query)` instead of `server.query(query)`). If the collection is not specified Solr will simply use the default collection name set on the SolrServer that you're operating on. These code changes not only update all of our SolrJ API calls to specify a collection but also method signatures were modified to allow sending a collection name and `SolrConfiguration` was updated to have methods to call to retrieve the collection we want to work on when we're doing a query/full index/incremental index. Because this logic was created in `SolrConfiguration` it left `BroadleafCloudSolrClient` unneeded since the only reason it existed was to override the `getDefaultCollection` in `CloudSolrClient`. We couldn't simply leave all of our code the same because even though it would make sense that Solr would call `getDefaultCollection` in order to find the default collection if it was not specified it in fact does not and instead refers to the property directly. Since this left our override pointless it made sense to simply move that logic into `SolrConfiguration` so that implementors were no longer forced to use our custom `SolrClient` if they were in a Solr Cloud environment. Additionally all public interface methods that needed the "collection" parameter added were left and deprecated with implementations that delegate to the method with the "collection" parameter in order to satisfy backwards compatibility. The downside to simply leaving it up to Solr to determine the default collection is in the event you want every site in a MT solution to have its own collection we no longer can override which collection we should be operating on so it will simply always use the default collection.

This PR solves https://github.com/BroadleafCommerce/QA/issues/3493

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/search/service/solr/index/SolrIndexServiceImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/search/service/solr/index/SolrIndexServiceImpl.java
@@ -317,7 +317,7 @@ public class SolrIndexServiceImpl implements SolrIndexService {
             String deleteQuery = StringUtil.sanitize(shs.getNamespaceFieldName()) + ":(\"" 
                     + StringUtil.sanitize(solrConfiguration.getNamespace()) + "\")";
             LOG.debug("Deleting by query: " + deleteQuery);
-            server.deleteByQuery(deleteQuery);
+            server.deleteByQuery(collection, deleteQuery);
 
             //Explicitly do a hard commit here since we just deleted the entire index
             server.commit(collection);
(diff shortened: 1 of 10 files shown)


ncrum (approved):
@cja769 this looks good as far as I can tell!

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/search/service/solr/index/SolrIndexServiceImpl.java
```diff
@@ -746,7 +746,7 @@ public void commit(SolrClient server, boolean softCommit, boolean waitSearcher,
                         + ", waitSearcher: " + waitSearcher + ", waitFlush: " + waitFlush);
             }
 
-            server.commit(waitFlush, waitSearcher, softCommit);
+            server.commit(collection, waitFlush, waitSearcher, softCommit);
```
phillipuniverse, line 749: This works if collection is `null`? How does Solr interpret this?

cja769: It uses the default collection for the server you're using. If you call the `commit` without the collection it calls the `commit` with the collection and sends null.

phillipuniverse (approved):
LGTM

PR #2021, opened 2018-11-29
Disallow customers from logging in if they're archived

cja769:
Since `Customer`s aren't sandboxabled the Hibernate filter to filter out archived `Customer`s isn't weaved on therefore in the `UserDetailsServiceImpl` we allow a `Customer` to log in even if they've been archived.

### DIFF
--- a/core/broadleaf-profile/src/main/java/org/broadleafcommerce/profile/core/service/UserDetailsServiceImpl.java
+++ b/core/broadleaf-profile/src/main/java/org/broadleafcommerce/profile/core/service/UserDetailsServiceImpl.java
@@ -17,6 +17,7 @@
  */
 package org.broadleafcommerce.profile.core.service;
 
+import org.broadleafcommerce.common.persistence.Status;
 import org.broadleafcommerce.profile.core.domain.Customer;
 import org.broadleafcommerce.profile.core.domain.CustomerRole;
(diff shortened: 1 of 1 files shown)


cja769:
@phillipuniverse I'm assuming 5.2 is far enough to backport but if not I can rebase it on top of a different line.

phillipuniverse (changes requested):
Archive status is weaved on in Enterprise (right? or did we move this into core?), you cannot just straight up cast to `Status`. Pretty sure this will blow up in community 💣💣💣💣💣💣💣💣

cja769:
@phillipuniverse As you see in 
screen shot 2018-12-06 at 10 52 17 am
 `DirectCopyTransformer` is actually in common and the archiveOnly token is hooked up. I can add a check anyways just in case someone excludes `Customer` from being archivedOnly

phillipuniverse:
👍 got it. Once you verify that login works with community then we're good to merge.

phillipuniverse:
Actually, I honestly think this is a p easy defensive coding measure anyway; this could be a class that has other tendrils that people are using in different contexts (without bytecode weaving).

It doesn't really hurt anything to add an `if (Status.isAssignableFrom(customer.getClass()))` to this, so go ahead and do that.

PR #2022, opened 2018-11-29
When validation fails, copy over any GlobalValdiationErrors

JustinCRL:
Currently when a validation error occurs a new entity retrieved, and the property validation errors are copied over.  However, any global validation errors that were created during the validation process is not copied over to the new instance.  This change would allow global validation errors to be displayed.


JustinCRL:
I will be submitting a separate PR for adding the ability to do Entity/Form Bean validation.  This framework requires this code change, so I will just include it with that commit.

PR #2024, opened 2018-12-03
Adding listener to OrderItemImpl and null check for tests

riteshadhikari17:
https://github.com/BroadleafCommerce/QA/issues/3502

https://secure.helpscout.net/conversation/694365208/31254?folderId=1819320

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/audit/AbstractAuditableListener.java
+++ b/common/src/main/java/org/broadleafcommerce/common/audit/AbstractAuditableListener.java
@@ -94,10 +94,10 @@ public abstract class AbstractAuditableListener {
                         auditable = field.get(entity);
                     }
 
-                Field temporalField = auditable.getClass().getDeclaredField(dateField);
-                Field agentField = auditable.getClass().getDeclaredField(userField);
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/broadleafcommerce/common/audit/AbstractAuditableListener.java
```diff
@@ -85,17 +85,20 @@ protected void setAuditUpdateData(Object entity, Object auditableObject) throws
...
+            if (field != null) {
+                field.setAccessible(true);
+                if (field.isAnnotationPresent(Embedded.class)) {
+                    Object auditable = field.get(entity);
+                    if (auditable == null) {
+                        field.set(entity, auditableObject);
+                        auditable = field.get(entity);
+                    }
+
                 Field temporalField = auditable.getClass().getDeclaredField(dateField);
```
Jerry77oz: @riteshadhikari17 can you indent these? it looks like they are misaligned with the if statement that it is in.

Jerry77oz (changes requested):
@riteshadhikari17 take a look at the inline comment. It's a small adjustment.

Jerry77oz (approved):

PR #2025, opened 2018-12-03
BroadleafEntityValidator enables entity/form validation

JustinCRL:
- If you extend BroadleafEntityValidator and specify the Hibernate entity as the generic, you can add custom validation having access to the full entity values prior to insert or update.
- When validation fails, copy over any GlobalValdiationErrors.  This allows BroadleafEntityValidator to create global errors, not just field errors.

As an example you could add product validation like this

    @Component
    public class AdminUserEntityValidator extends BroadleafEntityValidator<AdminUserImpl> {
  
      @Override
      void validate(Entity submittedEntity, AdminUserImpl adminUser, Map<String, FieldMetadata> propertiesMetadata, RecordHelper recordHelper, boolean validateUnsubmittedProperties) {
      
        //Skip validation if basic validation fields (e.g. required fields)
        if(!submittedEntity.isValidationFailure()) {
          if(adminUser.getLogin().equals("admin")) { 
            submittedEntity.addValidationError("login", "User name must not be 'admin'");
          }
        
          boolean similarUserExists = //Query the database with an injected DAO
          if(similarUserExists) {
            submittedEntity.addGlobalValidationError("A user already exists with...);
          }
        }
      }
    }

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/BasicPersistenceModule.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/BasicPersistenceModule.java
@@ -409,6 +409,7 @@ public class BasicPersistenceModule implements PersistenceModule, RecordHelper, 
                 entityList.add(instance);
                 Entity invalid = getRecords(mergedProperties, entityList, null, null, null)[0];
                 invalid.setPropertyValidationErrors(entity.getPropertyValidationErrors());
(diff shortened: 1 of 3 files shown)


JustinCRL:
https://build.broadleafcommerce.com seems unreachable.  Could you tigger the CR checks when it's back up.

JustinCRL:
If it's important to handle any part of the object hierarchy (e.g. referencing the interface as opposed to the implementation), you could lazy load a map of the objects passed in and whether any matches were found using isAssignableFrom.  However, this would be much slower than this implementation.

PR #2044, opened 2018-12-26
Was not able to add new translations for CustomField FriendlyName field

danielcolgrove:
When adding a CustomField translation the assignable (parent) class (Field) was not being used as the TranslatedEntity class.  Added ability to find assignable translated entity class to the TranslationService.  Using that new method from the AminTranslationController.  Addresses https://github.com/BroadleafCommerce/QA/issues/3547


StanislavFedorov (approved):
I have tested. Instead of CustomField class we got Field class and we have not exception when we adding translation.

PR #2048, opened 2019-01-03
BroadleafCommerce/QA#3539 - Fix to OrderItemPriceDetail not syncd from OrderItem when OrderItem.isDiscountingAllowed is false

jetchegoyen:
QA [#3539](https://github.com/BroadleafCommerce/QA/issues/3539)


jetchegoyen:
I tested with a 50% discount offer adding to the cart two items that apply to that offer.
One of the order items had  isDiscountingAllowed = false.  

image


image

As you can see in above screenshots, the discount was applied only to the proper item, and the price detail of the item that has discounting not allowed is updated on quantity modifications.

PR #2054, opened 2019-01-25
Added new interface method, that will returns all site parent ids.

Vitalij-Voronkoff:
BroadleafCommerce/QA#3572


common/src/main/java/org/broadleafcommerce/common/file/service/BroadleafFileServiceExtensionHandler.java
```diff
@@ -28,4 +31,6 @@
 
     public ExtensionResultStatusType processPathForSite(String prefix, String resourceName, ExtensionResultHolder<String> holder);
 
+    public ExtensionResultStatusType retrieveAllSiteIds(Site site, ExtensionResultHolder<List<String>> holder);
```
oleksiimiroshnyk: maybe rename method to parent sites ids or something like that ?

Vitalij-Voronkoff: yes, the name should be more clear

PR #2067, opened 2019-02-26
Added readOnly attribute validation because Offer Restriction Rules do not honor admin privileges

tolyshevchuk:
BroadleafCommerce/QA#3570
Offer Restriction Rules do not honor admin privileges. Added readOnly attribute validation.


dmyroniuk (approved):
It looks good.

PR #2069, opened 2019-02-27
Use the currency of the order instead of the default currency in checkout

cja769:
In the `ValidateAndConfirmPaymentActivity` we need to add the payments up based on the order's currency instead of the default currency since the default/current currency/locale isn't always the same as the order. This is a possible scenario when using the API since there's no sessions and certain payment gateways use redirects which may not allow the locale/currency to be set on the request

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/checkout/service/workflow/ValidateAndConfirmPaymentActivity.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/checkout/service/workflow/ValidateAndConfirmPaymentActivity.java
@@ -247,7 +247,7 @@ public class ValidateAndConfirmPaymentActivity extends BaseActivity<ProcessConte
         // Add authorize and authorize_and_capture transactions;
         // there should only be one or the other in the payment
         // Also add any pending transactions (as these are marked as being AUTH or CAPTURED later)
-        Money paymentSum = new Money(BigDecimal.ZERO);
+        Money paymentSum = new Money(BigDecimal.ZERO, order.getCurrency());
         for (OrderPayment payment : order.getPayments()) {
             if (payment.isActive()) {
                 paymentSum = paymentSum.add(payment.getSuccessfulTransactionAmountForType(PaymentTransactionType.AUTHORIZE))


phillipuniverse (approved):

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/checkout/service/workflow/ValidateAndConfirmPaymentActivity.java
```diff
@@ -247,7 +247,7 @@ public ValidateAndConfirmPaymentActivity(@Qualifier("blConfirmPaymentsRollbackHa
         // Add authorize and authorize_and_capture transactions;
         // there should only be one or the other in the payment
         // Also add any pending transactions (as these are marked as being AUTH or CAPTURED later)
-        Money paymentSum = new Money(BigDecimal.ZERO);
+        Money paymentSum = new Money(BigDecimal.ZERO, order.getCurrency());
```
phillipuniverse, line 250: Just FYI, another fix to this would have been to use `Money.ZERO` as a constant instead, see https://github.com/BroadleafCommerce/QA/issues/3583

cja769: Using `Money.ZERO` would result in the same issue that we're facing because that uses the default currency as well instead of the currency of the order. The issue is if the default currency is `USD` but the payment/order are in `AED` then the payments get added up in `USD` instead of `AED` since `Money#add` will use the "left-side" of the expression for the currency.

phillipuniverse: Ah yes, but you missed the money (heh) quote from the linked slack thread in that issue:

> the if statement is designed to say “if this is the literal `Money.ZERO` value directly here (not a `Money` who happens to have a value of zero, which is the case for `new Money(BigDecimal.ZERO)`), then ignore all the currency of _this_ Money, and just take on the currency of the one that is being added to me.

this is in the add() method of `Money`:

https://github.com/BroadleafCommerce/BroadleafCommerce/blob/9c1935ee9fb7eebefb3e972c7e887cfe1e28e2ed/common/src/main/java/org/broadleafcommerce/common/money/Money.java#L178-L183

whenever you do any operations with the _literal_ `Money.ZERO`, the currency established on that (default currency) is 100% ignored.

cja769: Unless you have a currency conversion service. I guess the question then is should we ignore the currency conversion service if it is `Money.ZERO` or should the currency conversion service handle that case?

phillipuniverse: > Unless you have a currency conversion service. I guess the question then is should we ignore the currency conversion service if it is `Money.ZERO` or should the currency conversion service handle that case?

~~That's the point, the `add()` and `subtract()` methods _completely ignore_ the currency of `Money.ZERO` and always results in the currency of the operand that is _not_ `Money.ZERO`. No `CurrencyConversion` is necessary, because the currency of `Money.ZERO` is completely ignored.~~

~~That is the meat of the code snippet that I linked.~~

I missed this comment from you:

> Unless you have a currency conversion service

You're right. In this case it would convert the currency when it shouldn't.

phillipuniverse: Ok I concede, the fix you have is the right one. Arguably yes we should ignore the `Money.ZERO` currency _in all cases_. But the idea of a `CurrencyConversionService` is both an esoteric override case and not one that I would _ever_ recommend anyone do (Broadleaf should really never do currency conversion).

But because it's in `Money` it's a risky change to make and difficult to predict the true ramifications.

All-in-all, you made the right decision A+++

cja769: I agree. I think that `Money#add` and `Money#subtract` should be modified to ignore the conversion service if it's `Money.ZERO` but I like to avoid changing `Money` like a plague.

PR #2073, opened 2019-03-05
The Admin datepicker requires the user to hit enter after typing in a…

tolyshevchuk:
The Admin datepicker requires the user to hit enter after typing in a date-time manually for it to be saved.
Added validation of manual input date and date from datepicker.


Jerry77oz (approved):

Jerry77oz:
@tolyshevchuk when we have a QA issue, add it on here so it get's referenced back. For example:

BroadleafCommerce/QA#3519

PR #2081, opened 2019-03-14
Changing addresses from a taxable address to a non-taxable address still retains original tax

cja769:
The issue was caused by checking if the tax factor was zero or null and then skipping all subsequent operations. Because of this if there was already a non-zero tax applied it wouldn't be removed when the tax was changed to zero. The core logic around this was moved to `applyTaxFactor` where now it handles looking for the case where if a tax record is found and the tax factor to be applied is zero then the tax record is removed from the list of tax details.

Additionally `calculateTaxForOrder` method was heavily refactored since it had repetitive code and was hard to modify in a client application because of its size.


phillipuniverse (approved):
👍 LGTM, your reasoning and what you did by removing existing tax records makes sense. I guess the only thing that we support is the `COMBINED` tax anyway?

PR #2082, opened 2019-03-20
Fixed an issue with internalization messages locale filter

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/3606


Jerry77oz:
@StanislavFedorov the fix for the actual filter UI is much better now but we are still hitting a small issue. We now have the filter that uses `contains`

image

But setting `en` in the field returns no results. It looks like it is doing an exact match since if I enter in `en_US` it does bring back the messages that have that locale.

Jerry77oz (changes requested):
See comment in conversation.

StanislavFedorov:
@Jerry77oz 
I have solved this issue in other way. Via "is in" filter (but not "contains"). I have reworked a little bit both backend and frontend code.

Jerry77oz (approved):

Jerry77oz:
Looks good!

PR #2090, opened 2019-04-02
Upgrade woodstox core asl version

Vitalij-Voronkoff:
Requires:BroadleafCommerce/BroadleafCommerce#2089

This upgrade needs to fix conflict old libraries needed to Apache CXF. BroadleafCommerce/blc-cybersource-payment#15

### DIFF
--- a/common/src/test/java/org/broadleafcommerce/common/rule/MvelHelperTest.java
+++ b/common/src/test/java/org/broadleafcommerce/common/rule/MvelHelperTest.java
@@ -112,45 +112,9 @@ public class MvelHelperTest extends TestCase {
         boolean result = MvelHelper.evaluateRule("request.properties['blcSearchTerm'] == 'hot'", parameters);
         assertTrue(result);
     }
-
-    /**
-     * Confirms MVEL failure for special overloaded method case.
-     * </p>
-     * During compilation, if an mvel expression contains a method calls with params, mvel will attempt to identify a perfect
-     * method signature match based on the types of the params. However, if the method is overloaded and one or more of
-     * the params are null, mvel will not be able to 100% identify the right method version if the overloaded methods
-     * have the same quantity of params and have type overlap. Take this example:
-     * </p>
-     * {@code SelectizeCollectionUtils#intersection(String param, Iterable param2);}
-     * {@code SelectizeCollectionUtils#intersection(Iterable param, Iterable param2);}
-     * </p>
-     * If a rule is executed calling the intersection method and a null is passed in for the first parameter, mvel will
-     * not know which method to pick, so it will end up picking one of them. This has proven to be undetermined at runtime,
-     * which has made the problem even harder to identify. Furthermore, once the choice is made during compilation,
-     * that compiled expression will utilize the "incorrect" choice going forward, which will cause the expression to
-     * fail, even when neither param is null. In fact, in the case above, mvel will "coerce" the first param Iterable
-     * instance into a String (i.e. flatten the list representation into a comma delimited String) in order to continue
-     * calling the method identified during compilation. This causes the expression to permanently be in an incorrect
-     * state from which it will never recover.
-     */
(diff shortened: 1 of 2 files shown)


common/src/test/java/org/broadleafcommerce/common/rule/MvelHelperTest.java
```diff
@@ -112,45 +112,9 @@ public void testRequestMapProperty() {
...
-     * If a rule is executed calling the intersection method and a null is passed in for the first parameter, mvel will
-     * not know which method to pick, so it will end up picking one of them. This has proven to be undetermined at runtime,
-     * which has made the problem even harder to identify. Furthermore, once the choice is made during compilation,
-     * that compiled expression will utilize the "incorrect" choice going forward, which will cause the expression to
-     * fail, even when neither param is null. In fact, in the case above, mvel will "coerce" the first param Iterable
-     * instance into a String (i.e. flatten the list representation into a comma delimited String) in order to continue
-     * calling the method identified during compilation. This causes the expression to permanently be in an incorrect
-     * state from which it will never recover.
-     */
-    public void testMvelMethodOverloadFailureCase() throws IOException {
```
phillipuniverse, line 136: Why did you delete this test?

pom.xml
```diff
@@ -1114,7 +1114,7 @@
             <dependency>
                 <groupId>org.mvel</groupId>
                 <artifactId>mvel2</artifactId>
-                <version>2.2.1.Final</version>
+                <version>2.4.3.Final</version>
```
phillipuniverse, line 1117: Why did you need to update mvel?

phillipuniverse (changes requested):
I don't really have the context to approve/deny this PR, but the MVEL change seems suspect.

common/src/test/java/org/broadleafcommerce/common/rule/MvelHelperTest.java
Vitalij-Voronkoff: this test failures while project building. I found that in 5.2 this test was deleted as not necessary because mvel version upgraded. I need to build framework version 5.0 because I need to upgrade woodstox library version, because old version causes error working with Apache CXF

Vitalij-Voronkoff: Context this task https://github.com/BroadleafCommerce/blc-cybersource-payment/pull/15

riteshadhikari17 (approved):

phillipuniverse:
@riteshadhikari17 it looks like the questions that I had and the reason that I requested changes went unanswered. What is the reason for us upgrading this version of MVEL in this 5.0 version?

danielcolgrove:
@Vitalij-Voronkoff has been on vacation.  I suspect it's a dependency caused by the upgrade from Axis1 to Axis2.

riteshadhikari17:
@phillipuniverse Its actually upgrade from axis1 to cxf. Vitali has notes on this PR https://github.com/BroadleafCommerce/blc-cybersource-payment/pull/15

danielcolgrove:
Yes, correct....not axis2.

phillipuniverse:
@danielcolgrove @riteshadhikari17 I think you guys are missing the point, or maybe I am? This PR includes an upgrade to MVEL. Why does this PR have anything to do with MVEL? What does MVEL have to do at all with axis or cxf?

If the issue was a test failure _without_ upgrading the MVEL version, then it sounds like that test failure was legitimate.

danielcolgrove:
@phillipuniverse I understand the question.  I would need to defer to Vitalii as to why it was required.  Upgrading MVEL wasn't part of the original unit of work.  I suspect the upgrade came in as a series of dependency upgrades.  Vitalii is back toward the end of the week and we can get more details once he returns.  @riteshadhikari17 If it makes sense from a timing perspective to revert this issue, let's do that.

PR #2110, opened 2019-05-16
Upgrade jQuery to latest version 3.4.1

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/3656

There were several type of issues:

1. Deprecated event aliases. 

".load, .unload, and .error, deprecated since jQuery 1.8, are no more. Use .on() to register listeners"
in files:
blc-admin.js, new-admin.js, oms-admin.js

2. Current version of jquery.event.move.js did not support 3.4.1 jQuery

I have upgrated it to newest version 2.0.0

3. Datapicker had broken
This commit https://github.com/BroadleafCommerce/BroadleafCommerce/commit/808f224b953cd3c8a541fde07709de00892e41c3 to jquery.datetimepicker.js had code that did not support 3.4.1 jQuery 

Fixed

4. Current version of redactor.js did not support 3.4.1 jQuery

Now redactor.js is commercial project https://imperavi.com/redactor/ . So i just fixed code that cause an error.


cja769:
@riteshadhikari17 @StanislavFedorov This needs to be reverted. Because of this PR lazy loading (at least for ToOne lookups though I'm sure other places) no longer work. I believe it's because the mCustomScrollbar jQuery library we use needs to be upgraded as well. In general I think there's a lot more work to do with this upgrade than what's been done. We have around 30 third party jQuery plugins that we use that will all likely need to be upgraded. These changes are really dangerous considering this is in a patch release of 5.2 and we use jQuery so heavily. Below is a gif of the error that I'm seeing
https://cl.ly/8c1d2f208855
these are just the plugins prefixed with `jquery` which is probably not all of them to put in perspective how many jQuery plugins we rely on
https://cl.ly/8c7c8ef72608

StanislavFedorov:
@cja769 
The main problem with this upgrade in that we have a huge gap between JQuery version that we had before and actual.
When we started this process i mentioned that this is complicated and time cunsuming update. And in fact we have 2 ways.
1. To make this update and to fix (or updates) our Jquery components 
2. To have a separate project related with "JQuery update". Because, in my opinion, it is not a level of QA issue and 1 developer. In fact somethimes it difficult to predict how these "third party" components will work with latest version of JQuery.

And it seems that we have choosen 1st way. Because of old version JQuery that we had, have a lot of vulnerabilities and upgrade have been marked as critical issue.

And the next issue with this update that In fact we already have made a lot of fixes after this first PR

riteshadhikari17:
@cja769 Almost all lib have dependency of jQuery 1.x or above(like 1.6 or 1.7 or 1.8 plus). None of them have any information for 3.x  I am going through all the jQuery plugins and making sure everything works as expected and  so far it looks good. I am going to take one more day to test everything and also merge everything to 6.0 and make sure 6.0 is also fine. If I don't find anything, let's not revert and fix issues as we find them. Because it doesn't look like its breaking a lot and 1 or 2 bugs can be hotfix until next release which we do every month.
@StanislavFedorov Could you look at the CustomScroll issue and fix that?

PR #2114, opened 2019-05-22
Fixed missing add button in global admin.

Vitalij-Voronkoff:
BroadleafCommerce/QA#3664

The problem can be occur when using/switching to jQuery 3. It's because all ready states in the new jQuery 3 are now fully asynchron. This means, that there is no given order for your code to be executed.

Because of this, it could happen, that load is been triggered before your ready state has been executed. When your ready function now finally gets triggered, your load listener is too late and will not be executed.

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/blc-admin.js
+++ b/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/blc-admin.js
@@ -1382,22 +1382,26 @@ var getCurrentHashVal = function() {
     return hash.substr(1);
 };
 
-function enableEntityButtonsIfReady() {
-// primary entity buttons should be disabled until page is loaded
+function finalPageLoadIfReady() {
     if (BLC.readyEventTriggered && BLC.loadEventTriggered) {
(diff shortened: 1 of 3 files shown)


admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/blc-admin.js
```diff
@@ -1380,12 +1380,12 @@ var getCurrentHashVal = function() {
     return hash.substr(1);
 };
 
-$(document).ready(function() {
-    // primary entity buttons should be disabled until page is loaded
-    $(window).on('load', function () {
-        $('.button.primary.large:not(.submit-button):not(.modify-production-inventory)').prop('disabled', false).removeClass('disabled');
-    });
+// primary entity buttons should be disabled until page is loaded
+$(window).on('load', function () {
```
oleksiimiroshnyk: is this the only place ?
I think there might couple of more places where we use onload handler, can you please doublecheck them also ?

Vitalij-Voronkoff: other places fixed in [OMS](https://github.com/BroadleafCommerce/OMS/pull/383)

PR #2121, opened 2019-06-10
Feature rating summary - total - last update

egorbatik:
**A Brief Overview**
RatingSummary as is, does not include a totalratings and the date where the rating were added, since the entity contains the average is common to have the total rating in this entity as well.

This has impact in the performance as well as secondary-inheriting entity doesn't have the same performance regarding caches.
 
**Additional context**
This came out in the context of the performance enhanctment that are being done in gopher.

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/rating/domain/RatingSummary.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/rating/domain/RatingSummary.java
@@ -17,9 +17,10 @@
  */
 package org.broadleafcommerce.core.rating.domain;
 
-import org.broadleafcommerce.core.rating.service.type.RatingType;
-
+import java.util.Date;
 import java.util.List;
+
+import org.broadleafcommerce.core.rating.service.type.RatingType;
 
 public interface RatingSummary {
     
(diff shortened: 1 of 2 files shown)


elbertbautista:
After talking with @danielcolgrove - since there are schema updates, we can't automatically merge this into the 5.2.x line. We need to re-factor these properties to be included as an optional weave via a system property

See the following in `bl-framework-applicationContext.xml` as an example:

```
    <bean id="blOfferUsesConditionalDirectCopyTransformerMember" class="org.broadleafcommerce.common.weave.ConditionalDirectCopyTransformMemberDto">
        <property name="templateTokens">
            <util:constant static-field="org.broadleafcommerce.core.offer.service.type.OfferOptionalDirectCopyTransformTypes.OPTIONAL_OFFER_USES_FIELDS"/>
        </property>
        <property name="conditionalProperty" value="enable.optional.offer.uses.fields"/>
    </bean>

    <bean id="blSkuMinOrderQuantityConditionalDirectCopyTransformerMember" class="org.broadleafcommerce.common.weave.ConditionalDirectCopyTransformMemberDto">
        <property name="templateTokens" value="optionalSkuMinOrderQuantityField"/>
        <property name="conditionalProperty" value="enable.sku.minOrderQuantity.field"/>
    </bean>
```

PR #2125, opened 2019-06-14
Allows not initial resource bundle requests to not specify mapping prefix and file list.

marieStandeven:
Resolves https://github.com/BroadleafCommerce/QA/issues/3686

Non initial bundle requests were failing due to file list and mapping prefix not being specified. Since they are not the initial request, they do not need to specify these attributes as noted in the Javadocs.
These changes allow for the mapping prefix and file list to be null - not specified - on not initial resource bundle requests.


phillipuniverse (approved):
Nice, LGTM

Issue #1292, opened 2015-03-26
Add api for resolving profile information from an entity


bpolster:
It is sometimes helpful for modules running in an "BLC MultiTenant" installation to easily have access to the underlying profile and catalog information.  

This API is not intended for use in non-BLC multi-tenant or enterprise installations.

PR #2130, opened 2019-06-24
Added property to allow sorting on a Product's parent Category to be disabled in the Admin due to unresolvable issues.

nathandmoore:
For BroadleafCommerce/QA#1292

To disable sorting, set `allow.product.parent.category.sorting=false`. In order not to cause unexpected changes in client environments, the default is `true`.


nathandmoore:
These commits are cherry-picked from 6.0. I'm pretty sure the best thing to do is to rebase and merge rather than squash or do a regular merge.

PR #2135, opened 2019-07-10
Don't save if validation error in dynamic form

oleksiimiroshnyk:
BroadleafCommerce/QA#3321
Don't save if validation error in dynamic form


oleksiimiroshnyk:
wrong soluition

PR #2147, opened 2019-08-07
Fix Solr alphabetical sort doesn't consider word order

marieStandeven:
**A Brief Overview**
When sorting A-Z on the product list page, the items don't appear to be sorted. However, it is sorting based on the lowest alphabetical word in the product name, and not using the product name as a phrase.
This PR adds the string field_type to the product name field to ensure that the products are sorted as a customer or business user would expect.

**Link to QA issue**
https://github.com/BroadleafCommerce/QA/issues/3726


phillipuniverse:
When does the sort actually get applied? I would expect that there would need to be some other change to sort on `product_name_s` instead of whatever it's doing now

marieStandeven:
@phillipuniverse actually, I had assumed that there was some sort of field type priority in Solr, and based on the correct sorting being sent back I had figured string had the higher priority. I know I shouldn't assume so I went back to verify and there is this property:

```
# These are the default sortable field types, if a client has additional field types
# to sort by, redefine this list with those types included
solr.sortable.field.types=sort,s,p,i,l
```

So, if any of the above field types exists on the field, it will use that for sorting.

phillipuniverse (approved):
👍 great, thanks for verifying that!

PR #2149, opened 2019-08-15
Fixed issue when field is not highlighting for adorned collection metadata

Vitalij-Voronkoff:
Fixed issue when field is not highlighting for adorned collection metadata, when defaultSku id do not match with product id.

BroadleafCommerce/QA#3409

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/dto/ParentRecordStructure.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/dto/ParentRecordStructure.java
@@ -1,0 +1,101 @@
+/*
+ * #%L
+ * BroadleafCommerce Open Admin Platform
+ * %%
+ * Copyright (C) 2009 - 2019 Broadleaf Commerce
+ * %%
+ * Licensed under the Broadleaf Fair Use License Agreement, Version 1.0
+ * (the "Fair Use License" located  at http://license.broadleafcommerce.org/fair_use_license-1.0.txt)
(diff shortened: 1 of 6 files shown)


admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/dto/ParentRecordStructure.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.io.Serializable;
+import java.util.Arrays;
+
+public class ParentRecordStructure implements Serializable, PersistencePerspectiveItem {
+
+    private static final long serialVersionUID = 1L;
+
+    private Entity parentRecord;
+    private ClassMetadata mainMetadata;
```
phillipuniverse: I think you should call this `parentMetadata`. It should match the same type of name as `parentRecord`.

common/src/main/java/org/broadleafcommerce/common/presentation/client/PersistencePerspectiveItemType.java
```diff
@@ -25,5 +25,6 @@
 public enum PersistencePerspectiveItemType {
     FOREIGNKEY,
     ADORNEDTARGETLIST,
-    MAPSTRUCTURE
+    MAPSTRUCTURE,
+    PARENTRECORDSTRUCTURE
```
phillipuniverse, line 35: Can you add a comment here with what this is used for?

phillipuniverse (changes requested):
Just the naming/comment changes, other than that this looks good to me.

phillipuniverse (approved):

PR #2151, opened 2019-08-22
Fixed an issue where Error appears in validation URL's for XSS on Admin

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/3738


oleksiimiroshnyk:
@StanislavFedorov btw, maybe null check should be included ?

PR #2156, opened 2019-09-09
Override missed equals hashcode serialVersionUID

putrenkov:
**A Brief Overview**
In some classes missed equals hashcode serialVersionUID the implements Serializable

**Link to QA issue**
QA-3182 - https://github.com/BroadleafCommerce/QA/issues/3182


charchar:
@putrenkov 	Looks like the class import ordering is incorrect. Assuming you are using intellij, please import these settings to your IDE: https://github.com/BroadleafCommerce/CodeStyle/blob/master/src/main/resources/broadleaf-java.importorder. FYI the order should be `org`, `com`, `java`, `javax`.

putrenkov:
@charchar Thanks I will fix it

charchar:
@putrenkov After talking with @danielcolgrove we want to fix the scenarios where a `Serializable` class has one of `equals()` or `hashCode()` but not the other. I pushed a test to `integration` module that checks for this situation and fails on a build. Can you resolve the 4 failing classes? You should be able to see them by running the test `EqualsHashCodeMethodsTest` in your IDE. Thanks!

putrenkov:
@charchar Hi, I ran a test it was successful.

admin/broadleaf-contentmanagement-module/src/main/java/org/broadleafcommerce/cms/url/domain/URLHandlerImpl.java
```diff
@@ -45,6 +45,9 @@
 import javax.persistence.InheritanceType;
 import javax.persistence.Table;
 
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
```
charchar: @putrenkov Still seeing import ordering issues throughout PR.

charchar (changes requested):

admin/broadleaf-contentmanagement-module/src/main/java/org/broadleafcommerce/cms/url/domain/URLHandlerImpl.java
putrenkov: @charchar Hi, could you check, is it better?

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/rulebuilder/dto/DataWrapper.java
```diff
@@ -27,6 +27,8 @@
 import java.io.Serializable;
 import java.util.ArrayList;
 
+import org.apache.commons.lang3.builder.HashCodeBuilder;
```
charchar: @putrenkov This is out of order

charchar (changes requested):

nathandmoore (changes requested):
Need to change `obj.getClass() != getClass()` to `getClass().isAssignableFrom(obj.getClass)` in each `equals()`.

PR #2157, opened 2019-09-13
Category Tree View doesn't Scroll, check footer existance

oleksiimiroshnyk:
BroadleafCommerce/QA#3751
check if footer available, so not to corrupt height

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/listGrid-paginate.js
+++ b/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/listGrid-paginate.js
@@ -794,7 +794,7 @@
 
                 var footerOuterHeight = $wrapper.next('.listgrid-table-footer:visible').outerHeight();
                 if (typeof footerOuterHeight !== "undefined") {
-                    wrapperHeight -= wrapperHeight1;
+                    wrapperHeight -= footerOuterHeight;
                 }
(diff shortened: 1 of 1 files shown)


admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/listGrid-paginate.js
```diff
@@ -792,7 +792,10 @@
                     wrapperHeight = $tbody.closest('.select-group').outerHeight();
                 }
 
-                wrapperHeight -= $wrapper.next('.listgrid-table-footer:visible').outerHeight();
+                var footerOuterHeight = $wrapper.next('.listgrid-table-footer:visible').outerHeight();
+                if (typeof footerOuterHeight !== "undefined") {
+                    wrapperHeight -= wrapperHeight1;
```
charchar: `wrapperHeight1` is not defined anywhere. I assume that it should be `footerOuterHeight`, instead?

charchar (changes requested):

admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/listGrid-paginate.js
oleksiimiroshnyk: correct. thx for catching this

charchar (approved):

PR #2159, opened 2019-09-16
Added SKU availability verification on checkout

Vitalij-Voronkoff:
**A Brief Overview**
There are two scenarios of SKU availability which allow them to be added to cart while available, then made unavailable, but still able to be checked out.

**Link to QA ticket**
BroadleafCommerce/QA#3750

**Link to HelpScout**
https://secure.helpscout.net/conversation/951360574/37473

**Additional context**
@phillipuniverse @epenning 
In my opinion the best place for this validation is create new extension handler, 
that extends `AbstractValidateCheckoutActivityExtensionHandler` and will be used in `ValidateCheckoutActivity`.
For validation I used the logic from `CheckUpdateAvailabilityActivity`. It seems all working fine.
**Cons**:
I do not meet DRY principle and I created code duplication.
This is because I can not simple extend `AbstractCheckAvailabilityActivity` for my purposes, because these activities in the other workflows and belongs to the different process contexts.
I need to reuse `checkSkuAvailability` method from the `AbstractCheckAvailabilityActivity` class. 
Also I used the logic from `CheckUpdateAvailabilityActivity.execute()`.

Could you give an advice for best solution? Perhaps we could extract methods that I need in my handler to some new Helper class?


phillipuniverse (changes requested):
The problem here is that I relied on an abstract class rather than have a service with a `checkSkuAvailability()` method on it that could be used in any circumstance.

That's what I think I would do in this case; move `checkSkuAvailability` into another service that is injected by the existing abstract activities, while continuing to use the same method signature.

Example of what `AbstractCheckAvailabilityActivity` will look like:

```java
public abstract class AbstractCheckAvailabilityActivity extends BaseActivity<ProcessContext<CartOperationRequest>> {

    @Resource(name = "blInventoryService")
    protected ContextualInventoryService inventoryService;

    @Override
    public boolean shouldExecute(ProcessContext<CartOperationRequest> context) {
        Order order = context.getSeedData().getOrder();
        return order != null && !Objects.equals(order.getStatus(), OrderStatus.NAMED);
    }
    
    protected void checkSkuAvailability(Order order, Sku sku, Integer requestedQuantity) throws InventoryUnavailableException {
        inventoryService.checkSkuAvailability();        
    }
}
```

Then put the old implementation into `InventoryService`. Doesn't _have_ to be in `InventoryService`, but I think you get the idea. This also means there is zero migration to support the changes we made in the abstract activity.

Now that that's extracted out of that class hierarchy, we can then figure out what we should do with checkout. I think what you proposed as an additional extension handler is ok, but maybe we should instead just add another activity into the checkout workflow? I think that would be a bit more clear and easier to follow. I would order it just after the existing `ValidateCheckoutActivity`, like maybe at order 750.

Vitalij-Voronkoff:
@phillipuniverse solution reworked according to https://github.com/BroadleafCommerce/QA/issues/3750#issuecomment-531783948

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/checkout/service/workflow/ValidateAvailabilityActivity.java
```diff
@@ -0,0 +1,86 @@
...
+                sku = ((BundleOrderItem) orderItem).getSku();
+            } else {
+                LOG.warn("Could not check availability; did not recognize passed-in item " + orderItem.getClass().getName());
+                return context;
+            }
+
+            Integer requestedQuantity = orderItem.getQuantity();
+            inventoryService.checkSkuAvailability(order, sku, requestedQuantity);
+
+            Integer previousQty = orderItem.getQuantity();
```
epenning: I believe this whole block following `inventoryService.checkSkuAvailability(order, sku, requestedQuantity);` is unnecessary, since `order.getOrderItems()` will include the child order items as well.

This block of code is logic duplicated from `CheckUpdateAvailabilityActivity`, the original block from `CheckUpdateAvailablilityActivity` is this:

```java
Order order = context.getSeedData().getOrder();
Integer requestedQuantity = request.getItemRequest().getQuantity();
checkSkuAvailability(order, sku, requestedQuantity);

Integer previousQty = orderItem.getQuantity();
for (OrderItem child : orderItem.getChildOrderItems()) {
    Sku childSku = ((DiscreteOrderItem) child).getSku();
    Integer childQuantity = child.getQuantity();
    childQuantity = childQuantity / previousQty;
    checkSkuAvailability(order, childSku, childQuantity * requestedQuantity);
}
```

In this usage for `CheckUpdateAvailabilityActivity`, `requestedQuantity` is a new, different value from `previousQty`, because the quantity is being _updated_. Therefore, the Sku availability for the newly scaled child quantity must be validated.

However, in `ValidateAvailabilityActivity`, the quantity is _not_ being updated, so `requestedQuantity` and `previousQty` are equal, so nothing is gained by the calculation of child quantity using both values. Therefore it shouldn't be a problem to remove this block.

epenning (changes requested):
One of the cases in the original issue was also that checkout was possible with an inactive sku/product, which doesn't seem to be covered here:

1. Create a new cart
2. Add to cart with a SKU while the date range is active
3. Update the SKU active date range to mark the SKU as inactive
4. Complete checkout successfully

It seems like checkout should also validate whether a SKU is inactive to confirm availability, unless there's any design decision to dictate otherwise.

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/checkout/service/workflow/ValidateAvailabilityActivity.java
Vitalij-Voronkoff: @epenning yes, you are right, this block is not needed. Also added additional active check.

epenning (approved):
Looks good to me

phillipuniverse (approved):

phillipuniverse:
Love the addition of the test! Great work!

PR #2164, opened 2019-09-20
remove PaymentTransaction from purge

putrenkov:
**A Brief Overview**
Delete order without PaymentTransaction for save history.

**Issue Link**
QA-3405 - https://github.com/BroadleafCommerce/QA/issues/3405


danielcolgrove:
@putrenkov I don't think we need this change.  The OrderDaoImpl#delete(Order) already accounts for PaymentTransactions by archiving them.

PR #2166, opened 2019-09-24
Fix to remove excessive calls to server when changing catalog/profile in admin

oleksiimiroshnyk:
BroadleafCommerce/QA#3761
fix for not refreshing page

### DIFF
--- a/common/src/main/resources/common_style/js/BLC.js
+++ b/common/src/main/resources/common_style/js/BLC.js
@@ -328,6 +328,37 @@ var BLC = (function($) {
             }
         }
 
+        return document.location.search = params;
+    }
+
+    /**
+     * Add a URL parameter (or changing it if it already exists)
+     * This method does the same as {@link #addUrlParam (search, key, val)},
+     * but does not update document.location.search to prevent page reload
+     * @param {search} string  this is typically document.location.search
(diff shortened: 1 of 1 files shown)


nathandmoore (changes requested):
@oleksiimiroshnyk Won't this break wherever `addUrlParam` is called and the caller expects the location to change? The function did what it did intentionally.

I would expect that we'd make a new function that has the new desired behavior and make sure to document the original one correctly. Then, update the `multitenant.js` where we use `BLC.addUrlParam` twice with the new function since that seems to be what was causing the issue described.

oleksiimiroshnyk:
@nathandmoore do you know who/what logic expects that adding url param will refresh a page?
I was thinking who and why can expect that and didn't found an answer, I tested the logic I could and it worked. Do you know if something is broken with this fix ?
Because to me it sounds like bad original impl.

nathandmoore:
@oleksiimiroshnyk The locale selector uses it, and we don't know what clients are using it in their code expecting the original behavior. Changing the fundamental behavior of the method could introduce bugs for them. We shouldn't do that in a patch release.

Vitalij-Voronkoff:
@nathandmoore solution reworked.

nathandmoore (approved):

PR #2169, opened 2019-09-27
Fixed an issue where Read only table still allows reorder edits

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/3183

I have removed drag-and-drop if listgrid. isReadOnly = true


nathandmoore (approved):
@StanislavFedorov This PR looks like it resolves the 1st of the 2 tasks specified:

> First, the readonly list grid should not allow reorders. Second, the system should fail the edit on the unowned profile.

Therefore, we should make sure not to mark BroadleafCommerce/QA#3183 as closed when this PR is merged.

PR #2170, opened 2019-09-30
NameResponse inner class without getters and Setters #3767

egorbatik:
The class
OrderToPaymentRequestDTOServiceImpl containes a inner class called NameResponse which contains firstName and lastName as protected fields.
The inner class has no getters and setters on it's memeber so when trying to extend that class there is no wahy to change the state of the inner class.

Just try to extend OrderToPaymentRequestDTOServiceImpl

Link to HelpScout
https://secure.helpscout.net/conversation/968506277/37884?folderId=2485392

As a workaround.... is possible to create a bean extending OrderToPaymentRequestDTOService, copying the the content of OrderToPaymentRequestDTOServiceImpl and changing the NameResponse innerclass.


riteshadhikari17:
Can't we add getters and setters?

PR #2176, opened 2019-10-10
Avoiding translation calls for the default locale

Vitalij-Voronkoff:
BroadleafCommerce/QA#3782

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/i18n/service/DynamicTranslationProvider.java
+++ b/common/src/main/java/org/broadleafcommerce/common/i18n/service/DynamicTranslationProvider.java
@@ -32,7 +32,6 @@ public class DynamicTranslationProvider {
     /**
      * If translations are enabled, this method will look for a translation for the specified field. If translations are
      * disabled or if this particular field did not have a translation, it will return back the defaultValue.
-     * If the passed in locale's language matches the Broadleaf default locale, it will return back the defaultValue.
      * 
      * @param obj
      * @param field
@@ -43,13 +42,11 @@ public class DynamicTranslationProvider {
         BroadleafRequestContext requestContext = BroadleafRequestContext.getBroadleafRequestContext();
         String valueToReturn = defaultValue;
 
-        // avoiding translation calls for the default locale
-        if (TranslationConsiderationContext.hasTranslation() &&
-                !requestContext.getLocale().getDefaultFlag()) {
+        if (TranslationConsiderationContext.hasTranslation()) {
             TranslationService translationService = TranslationConsiderationContext.getTranslationService();
             Locale locale = requestContext.getJavaLocale();
             String translatedValue = translationService.getTranslatedValue(obj, field, locale);
-            
+
             if (StringUtils.isNotBlank(translatedValue)) {
                 valueToReturn = translatedValue;
             } else {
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/broadleafcommerce/common/i18n/service/DynamicTranslationProvider.java
```diff
@@ -32,18 +32,22 @@
     /**
      * If translations are enabled, this method will look for a translation for the specified field. If translations are
      * disabled or if this particular field did not have a translation, it will return back the defaultValue.
+     * If the passed in locale's language matches the Broadleaf default locale, it will return back the defaultValue.
```
ckittrell:
```suggestion
     * If in the context of the default locale, then skip all translation calls and return the defaultValue.
```

ckittrell (approved):
Approving with the expectation that we cleanup the javadoc message

common/src/main/java/org/broadleafcommerce/common/i18n/service/DynamicTranslationProvider.java
```diff
@@ -32,18 +32,22 @@
...
      */
     public static String getValue(Object obj, String field, final String defaultValue) {
+        BroadleafRequestContext requestContext = BroadleafRequestContext.getBroadleafRequestContext();
         String valueToReturn = defaultValue;
-        
-        if (TranslationConsiderationContext.hasTranslation()) {
+
+        // avoiding translation calls for the default locale
+        if (TranslationConsiderationContext.hasTranslation() &&
+                !requestContext.getLocale().getDefaultFlag()) {
```
ckittrell: @Vitalij-Voronkoff On second thought, I think we should refactor this to be `&& localeShouldBeTranslated(requestContext.getLocale())`

Within that method, we can simply return `!locale.getDefaultFlag()`, but it would give users the ability to declare that `en_us` & `en_gb` should not be translated if only `en_us` is the default.

ckittrell (changes requested):

common/src/main/java/org/broadleafcommerce/common/i18n/service/DynamicTranslationProvider.java
ckittrell: ... realizing that `getValue(...)` is a static method which really derails the idea of extensibility

ckittrell (approved):

ckittrell:
@danielcolgrove @jfleschler I made a fair amount of changes on this one, so I'd like a second pair of eyes on it from one of you guys.

common/src/main/java/org/broadleafcommerce/common/i18n/service/TranslationServiceImpl.java
```diff
@@ -492,7 +510,7 @@ protected boolean propertyInDefaultLocaleExceptionList(Object entity, String pro
      * @param locale
      * @return
      */
-    protected boolean localeMatchesDefaultLocale(Locale locale) {
+    protected boolean languageMatchesDefaultLocaleLanguage(Locale locale) {
```
jfleschler: This is an API change that could break clients.  We probably don't want to have this as part of the PR

jfleschler (changes requested):

jfleschler (approved):
LGTM

Issue #2180, opened 2019-10-16
NullPointerException in BroadleafThemeProcessor when running site module 


frostyandy2k:
**Reproduce:**
```
https://github.com/BroadleafCommerce/DemoSite.git
cd DemoSite
git checkout tags/broadleaf-6.0.6-GA
mvn clean install
cd site/ && mvn spring-boot:run # navigate to the URL eg. `localhost:8080`
```

**Resulting Exception:**

```
2019-10-16 17:52:35.751 ERROR 19125 --- [nio-8443-exec-1] o.a.c.c.C.[.[.[/].[dispatcherServlet]    : Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception

java.lang.NullPointerException: null
	at org.broadleafcommerce.common.web.BroadleafThemeProcessor.process(BroadleafThemeProcessor.java:55) ~[broadleaf-common-6.0.6-GA.jar:na]
	at org.broadleafcommerce.common.web.BroadleafThemeResolverFilter.doFilterInternalUnlessIgnored(BroadleafThemeResolverFilter.java:96) ~[broadleaf-common-6.0.6-GA.jar:na]
	at org.broadleafcommerce.common.web.filter.AbstractIgnorableOncePerRequestFilter.doFilterInternal(AbstractIgnorableOncePerRequestFilter.java:58) ~[broadleaf-common-6.0.6-GA.jar:na]
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:107) ~[spring-web-5.0.8.RELEASE.jar:5.0.8.RELEASE]

...

	at org.apache.tomcat.util.net.SocketProcessorBase.run(SocketProcessorBase.java:49) [tomcat-embed-core-8.5.32.jar:8.5.32]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [na:1.8.0_222]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [na:1.8.0_222]
	at org.apache.tomcat.util.threads.TaskThread$WrappingRunnable.run(TaskThread.java:61) [tomcat-embed-core-8.5.32.jar:8.5.32]
	at java.lang.Thread.run(Thread.java:748) [na:1.8.0_222]

[...]

PR #2181, opened 2019-10-16
Fixes the issue where a ThemeDTO returns an uninitialized Object and crashes the site application #2180 uninitialized Long object causes NullPointerException

frostyandy2k:
**BUG**

NullPointerException due to a comparison on a null value in BroadleafThemeProcessor when running the site module (Issue: #2180).

The crash is due to [a comparison on the ID of a Null Object](https://github.com/BroadleafCommerce/BroadleafCommerce/blame/f2361835349b9ef2426d9c258b540f5613031d2e/common/src/main/java/org/broadleafcommerce/common/web/BroadleafThemeProcessor.java#L55) introduced as part of some optimization task.

This comes from the [uninitialized Long object in ThemeDTO](https://github.com/BroadleafCommerce/BroadleafCommerce/blame/f2361835349b9ef2426d9c258b540f5613031d2e/common/src/main/java/org/broadleafcommerce/common/site/domain/ThemeDTO.java#L28).

I guess this happens when running the project without any custom configuration of the Theme.

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/site/domain/ThemeDTO.java
+++ b/common/src/main/java/org/broadleafcommerce/common/site/domain/ThemeDTO.java
@@ -25,7 +25,7 @@ public class ThemeDTO implements Theme {
 
     public String path = "";
     public String name = "";
-    public Long id;
+    public Long id=0;
     
     public ThemeDTO() {
         // empty constructor


frostyandy2k:
The team is on it, closing this pull request in favor of this one: https://github.com/BroadleafCommerce/BroadleafCommerce/pull/2183

PR #2183, opened 2019-10-18
fix(theme): Made theme change detection null-safe

nathandmoore:
Resolves https://github.com/BroadleafCommerce/QA/issues/3793

**Issue**
Detecting a changed theme results in NPE if the original `ThemeDTO` was created by `NullBroadleafThemeResolver`.

**Fix**
Added null-safe comparison of 2 `ThemeDTOs` with the objects' `equals` and `hashcode` instead of checking the IDs since one of the DTOs may be empty (i.e., would have a `null` ID).

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/web/BroadleafThemeProcessor.java
+++ b/common/src/main/java/org/broadleafcommerce/common/web/BroadleafThemeProcessor.java
@@ -51,7 +51,7 @@ public class BroadleafThemeProcessor extends AbstractBroadleafWebRequestProcesso
         Theme newTheme = themeResolver.resolveTheme(request);
 
         //Track if the theme changed
-        if (originalTheme != null && newTheme != null && !Objects.equals(originalTheme, newTheme)) {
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/broadleafcommerce/common/web/BroadleafThemeProcessor.java
```diff
@@ -46,18 +46,16 @@ public void process(WebRequest request) {
...
         Theme newTheme = themeResolver.resolveTheme(request);
-        
+
         //Track if the theme changed
-        if (originalTheme != null && newTheme != null) {
-            if (!originalTheme.getId().equals(newTheme.getId())) {
-                Map<String, Object> properties = brc.getAdditionalProperties();
-                properties.put(BroadleafThemeResolver.BRC_THEME_CHANGE_STATUS, true);
-            }
+        if (!Objects.equals(originalTheme, newTheme)) {
```
nathandmoore:
```suggestion
        if (originalTheme != null && newTheme != null && !Objects.equals(originalTheme, newTheme)) {
```

PR #2188, opened 2019-10-31
Not able to add Sku with options to cart, hibernate proxy is not deproxied correctly

oleksiimiroshnyk:
BroadleafCommerce/QA#3808
Check for another proxy pattern

@jefffischer @phillipuniverse not sure what is real root cause of this issue, either another proxy lib is used or javassist upgrade or another lib upgrade but patter was changed. I didn't find any new props set to change this behavior
Not sure if this method has a good signature - it accepts class, would be better if object itself, in this case we can use deproxy methods from hibernate...


jefffischer:
This seems like it's always been brittle. Maybe something more along the lines of: https://stackoverflow.com/questions/2216547/how-to-convert-a-hibernate-proxy-to-a-real-entity-object

Specifically, ((HibernateProxy) entity).getHibernateLazyInitializer().getEntityName() to get the real entity class. I assume this is the goal at the end of the day.

PR #2189, opened 2019-11-01
fix error veiw price list remove

putrenkov:
**A Brief Overview**
If product have price list and when we try remove it throw exception.

**Link to QA issue**
QA-3765-https://github.com/BroadleafCommerce/QA/issues/3765


danielcolgrove (changes requested):
It appears you added some changes unrelated to this issue (order purge)

putrenkov:
@danielcolgrove Thanks Daniel, good catch.

PR #2190, opened 2019-11-04
Cannot upload an image from the Product page on community edition

jetchegoyen:
BroadleafCommerce/QA#3814

- changed to .on('load', fn) event handler for iframe in assetSelector.js instead of .load(fn). Since jquery 3.0, load fn is used for fetching data.
- Removed 'text/html' content type header from `upload` mapping as it produces 'application/json'

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/assetSelector.js
+++ b/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/assetSelector.js
@@ -36,11 +36,11 @@
                 'class' : 'upload_target hidden'
             });
             $form.before($iframe);
-            
+
             if (currentRedactor == null) {
-                $iframe.load(this.iframeOnLoad);
+                $iframe.on('load', this.iframeOnLoad);
             } else {
-                $iframe.load(this.iframeOnLoadRedactor);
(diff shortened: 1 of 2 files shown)


admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/assetSelector.js
```diff
@@ -36,11 +36,11 @@
...
             });
             $form.before($iframe);
-            
+
             if (currentRedactor == null) {
-                $iframe.load(this.iframeOnLoad);
+                $iframe.on('load', this.iframeOnLoad);
             } else {
-                $iframe.load(this.iframeOnLoadRedactor);
+                $iframe.on('load', this.iframeOnLoadRedactor);
```
oleksiimiroshnyk, line 43: @jetchegoyen just to doublecheck, did you test your change ?
I treied to do something like that and for me function was invoked before content is loaded. Maybe I did something wrong, so just to doublcheck

jetchegoyen: @oleksiimiroshnyk yes, I tested it and worked fine. The function was called once the iframe got loaded. 
I just checked again and worked fine.

PR #2194, opened 2019-11-12
Fixes issue where change details aren't captured for SKU modifications

jfleschler:
BroadleafCommerce/QA/issues/3833

**A Brief Overview**
Using the alternate transition pipeline, sku updates were not being captured when a sku had product option values associated with it.  Moving the merge about any option value processing ensures that the changes are merged and subsequently flushed separate from the removal of a product option value.

### DIFF
--- a/admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/server/service/handler/SkuCustomPersistenceHandler.java
+++ b/admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/server/service/handler/SkuCustomPersistenceHandler.java
@@ -18,8 +18,6 @@
 
 package org.broadleafcommerce.admin.server.service.handler;
 
-import static com.google.common.base.CharMatcher.DIGIT;
-
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.collections.Transformer;
 import org.apache.commons.lang.BooleanUtils;
@@ -37,7 +35,6 @@ import org.broadleafcommerce.common.presentation.client.SupportedFieldType;
 import org.broadleafcommerce.common.presentation.client.VisibilityEnum;
 import org.broadleafcommerce.common.sandbox.SandBoxHelper;
 import org.broadleafcommerce.common.util.BLCCollectionUtils;
-import org.broadleafcommerce.common.util.StringUtil;
 import org.broadleafcommerce.common.util.TypedTransformer;
 import org.broadleafcommerce.common.util.dao.DynamicDaoHelperImpl;
 import org.broadleafcommerce.core.catalog.domain.Product;
@@ -80,6 +77,7 @@ import org.broadleafcommerce.openadmin.server.service.persistence.module.criteri
 import org.hibernate.ejb.HibernateEntityManager;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
+
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -89,6 +87,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+
 import javax.annotation.Resource;
 import javax.persistence.EntityManager;
 import javax.persistence.PersistenceContext;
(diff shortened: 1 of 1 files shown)


admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/server/service/handler/SkuCustomPersistenceHandler.java
```diff
@@ -741,13 +740,16 @@ public Entity update(PersistencePackage persistencePackage, DynamicEntityDao dyn
                 return entity;
             }
 
+            // Persist our updates before making any other modifications.  Updating/removing
+            // ProductOptionValues cause an `em.flush()` downstream which can cause our changes
+            // to not be detected by the workflow management.
+            adminInstance = dynamicEntityDao.merge(adminInstance);
```
jefffischer, line 746: Curious why this is necessary. WorkflowEntityPersisterUpdateInterceptorExtensionHandler catches every flush and includes stuff in a ChangeDetailContext, which is eventually merged in SimpleChangeDetailDaoImpl. This should be capable of catching any persistence interaction(s) with the database during this flow and recording the change details.

jefffischer (changes requested):

admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/server/service/handler/SkuCustomPersistenceHandler.java
jfleschler: I believe it has to do with `AdminSandBoxEntityManagerInvocationProvider#handleRemoveState` which is hit when the `ProductOptionValues` are removed from the sku.  When the `WorkflowEntityPersisterUpdateInterceptorExtensionHandler` was being hit for the sku, it had `WorkflowStateManagerImpl.getWorkflowStateManager().getIgnoreUpdateInsertState()` set to true which caused it to skip the code that populates the `changesByTable` map.

Before swapping the order, `WorkflowEntityPersisterUpdateInterceptorExtensionHandler` was never hit when the merge was called for sku if it had product option values associated with it.  I'm guessing that the initial time through was because of a cascade from the remove but I'm not entirely sure.

jefffischer: I don't think one-off ordering changes are the right spot to try to deal with an idiosyncrasy of  WorkflowEntityPersisterUpdateInterceptorExtensionHandler. The design intent is that whatever is making it to the database during a flush (implicit or explicit) is interpreted by WorkflowEntityPersisterUpdateInterceptorExtensionHandler. It sounds like getIgnoreUpdateInsertState() caused that to fail in this case? If we try to think about this as messing with ordering in various spots, I don't think we'll ever get it right, especially across both strategies.

jefffischer: It may be that everything dirty was flushed during the product option value maintenance, and there was nothing to persist at the final merge (i.e. Hibernate thinks the level 1 cache state is the same as the persisted state).

jefffischer: Taking a quick look at the code in WorkflowEntityPersisterUpdateInterceptorExtensionHandler, I don't think the setupDetails methods should be gated with `!WorkflowStateManagerImpl.getWorkflowStateManager().getIgnoreUpdateInsertState()`.

jfleschler: That's what I was seeing.  I actually just pulled that out of the if statement and was going to run some tests to see what it affected.

jfleschler: @jefffischer I'm not seeing any immediate issues with removing the gating.  I'll close this PR and continue testing that solution.

PR #2196, opened 2019-11-13
Add special case when translating MVEL expressions for between expressions for numbers

cja769:
Currently when trying to use a `BETWEEN` or `BETWEEN_INCLUSIVE` expression an error is thrown saying that there's an invalid subgroup. This is caused because the MVEL expression is wrapped in parenthesis and subgroups aren't allowed for rule builders however this isn't actually a subgroup. Removing the parenthesis was too risky and may have unwanted side affects so instead an extra condition was added to a method that was also looking for an edge case around `BETWEEN` and `BETWEEN_INCLUSIVE` expressions when comparing dates. The condition simply checks if the operator is `BETWEEN` or `BETWEEN_INCLUSIVE` and if so then we deem it valid. Additionally tests were added to make sure we know if we have a regression in regards to this functionality in the future.

This PR fixes https://github.com/BroadleafCommerce/QA/issues/3829
Related to https://github.com/BroadleafCommerce/qa/issues/3070


jetchegoyen (commented):
I've just checked it and is working as expected

PR #2208, opened 2019-12-17
Fixed issue where multiple offers with minimum order subtotal requirements can get skipped

danielcolgrove:
Fixes https://github.com/BroadleafCommerce/QA/issues/3879


charchar (approved):
Looks good, makes sense to only add the potential if it's candidates still exist

Issue #1237, opened 2015-01-28
Cannot filter admin permissions in the admin


phillipuniverse:
Filtering permission lists (either when looking at a role or for an admin user) have zero effect.

PR #2210, opened 2019-12-20
Cannot see translations for profile entities in child site, that were added in parent

oleksiimiroshnyk:
BroadleafCommerce/QA#1237
fix for showing correct translations for profile entities(pages etc) on site


cja769:
This change can't go into 5.2 because it requires a data migration. The fix is going into 6.1 via https://github.com/BroadleafCommerce/BroadleafCommerce/pull/2211

PR #2215, opened 2019-12-31
Add a property to disable solr optimize call

oleksiimiroshnyk:
BroadleafCommerce/QA#3831
Add a property to disable solr optimize call

Property 
enable.solr.optimize

don't forget to add this prop to the doc

enable.solr.optimize false by default, checks if solr optimize will happen after index is rebuild. If you don't set this prop and somewhere you invoke SolrHelperService.optimize it will continue to work. To disable completely set it to false

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/search/service/solr/index/SolrIndexServiceImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/search/service/solr/index/SolrIndexServiceImpl.java
@@ -190,10 +190,10 @@ public class SolrIndexServiceImpl implements SolrIndexService {
         if(optimizeEnabled) {
             // this is required to be at the very very very end after rebuilding the whole index
             optimizeIndex(solrConfiguration.getReindexCollectionName(), solrConfiguration.getReindexServer());
-            // Swap the active and the reindex cores
-            if (!solrConfiguration.isSingleCoreMode()) {
-                shs.swapActiveCores(solrConfiguration);
-            }
+        }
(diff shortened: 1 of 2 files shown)


core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/search/service/solr/index/SolrIndexServiceImpl.java
```diff
@@ -183,11 +187,13 @@ public void buildIndex() throws IOException, ServiceException {
...
-        // this is required to be at the very very very end after rebuilding the whole index
-        optimizeIndex(solrConfiguration.getReindexCollectionName(), solrConfiguration.getReindexServer());
-        // Swap the active and the reindex cores
-        if (!solrConfiguration.isSingleCoreMode()) {
-            shs.swapActiveCores(solrConfiguration);
+        if(optimizeEnabled) {
+            // this is required to be at the very very very end after rebuilding the whole index
+            optimizeIndex(solrConfiguration.getReindexCollectionName(), solrConfiguration.getReindexServer());
+            // Swap the active and the reindex cores
+            if (!solrConfiguration.isSingleCoreMode()) {
```
danielcolgrove: Should the core swap be inside the optimizeEnabled block?  Seems like the core should be swapped regardless if optimize is enabled.

oleksiimiroshnyk: ah, yes, forgot that we when reindex we add to "reindex" core and swap should happen anyway

Issue #2227, opened 2020-01-07
Add Index on ProductImpl::urlKey to Accelerate Query


wtune:
Missing index on ProductImpl::urlKey may slow down the named query `BC_READ_PRODUCTS_BY_OUTGOING_URL`, which is used in `ProductDaoImpl::findProductByURI`.

PR #2229, opened 2020-01-07
Add index on ProductImpl::urlKey

wtune:
**A Brief Overview**
Add index on ProductImpl::urlKey can accelerate the named query `BC_READ_PRODUCTS_BY_OUTGOING_URL`. See #2227.

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/catalog/domain/ProductImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/catalog/domain/ProductImpl.java
@@ -108,12 +108,14 @@ import javax.persistence.Transient;
  */
 @Entity
 @Inheritance(strategy = InheritanceType.JOINED)
-@javax.persistence.Table(name = "BLC_PRODUCT", indexes = {@javax.persistence.Index(name = "IDX_BLPRD_URL_KEY", columnList = "URL_KEY")})
+@javax.persistence.Table(name = "BLC_PRODUCT")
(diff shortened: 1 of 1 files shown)


danielcolgrove:
The URL_KEY already have an index.  Not sure why the PR would not have recognized this.  See https://github.com/BroadleafCommerce/BroadleafCommerce/blob/develop-6.0.x/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/catalog/domain/ProductImpl.java.

This PR should be able to be closed.

wtune:
Thanks for response.

The existing index `PRODUCT_URL_INDEX` is a composite index on `(URL, URL_KEY)`. Queries with condition `URL=? AND URL_KEY=?` benefit from this index.

The named query `BC_READ_PRODUCTS_BY_OUTGOING_URL` mentioned in #2227 has condition `URL=? OR URL_KEY=?` though.  The composite index would not be used because it's not the leftmost prefix of the index ([MySQL reference](https://dev.mysql.com/doc/refman/8.0/en/multiple-column-indexes.html), [Postgres reference](https://www.postgresql.org/docs/12/indexes-multicolumn.html)). In such case, adding index on `URL_KEY` individually would allow MySQL to use [index merge](https://dev.mysql.com/doc/refman/8.0/en/index-merge-optimization.html) to accelarate the query. [Similar technique](https://www.postgresql.org/docs/8.3/indexes-bitmap-scans.html) is also available in Postgres. Thus, I believe the pull request was useful.

P.S. Changes in the PR was refactored to make it consistent to exisiting code in the class.

Issue #2228, opened 2020-01-07
Add Index on OfferImpl::startDate to Accelerate Query


wtune:
Missing index on OfferImpl::startDate may slow down query in `OfferDaoImpl::readOffersByAutomaticDeliveryType`.

PR #2230, opened 2020-01-07
Add index on OfferImpl::startDate

wtune:
**A Brief Overview**
Adding index on OfferImpl::startDate can accelerate the query in `OfferDaoImpl::readOffersByAutomaticDeliveryType`. See #2228.


danielcolgrove (changes requested):
For consistency in that class, can you add an @Index annotation on the field instead?

@Index(name="idx_BLOF_START_DATE", columnNames={"START_DATE"})

wtune:
Thanks for response! 

Now use `@Index` on the field instead.

PR #2231, opened 2020-01-07
Client performance solr reindex

ktisdell:
**Link to QA issue**
https://github.com/BroadleafCommerce/QA/issues/3922

**Link to FTD-specific installation instructions**
https://blc.slack.com/files/T02654EKZ/FS4C64H8B


core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/search/service/solr/indexer/AbstractSolrIndexUpdateServiceImpl.java
```diff
@@ -0,0 +1,292 @@
...
+        if (command != null) {
+            try {
+                for (int i = 0; i < 5; i++) {
+                    // Assuming there is not already an equal command in the queue, and not a FullReindexCommand in the queue, then add the command
+                    // for processing in a background thread.
+                    // If a FullReindexCommand is in the queue, then all incremental update commands are ignored since they'll
+                    // be overridden when the full reindex is applied. Just before the FullReindexCommand begins, it
+                    // is removed from the queue.  This allows incremental update commands to be queued with the assumption
+                    // that the full reindex may be mid-stream & may not have picked up the update.
+                    if (!commandQueue.contains(FullReindexCommand.DEFAULT_INSTANCE) && !commandQueue.contains(command)) {
```
ckittrell: @ktisdell I'm beginning to question this condition - do we really want to avoid queueing incremental update commands if a FullReindexCommand is in queue?

I think that logic is totally fine if the full reindex is successful, but if the full reindex fails, then you're left with a foreground collection that won't be updated.

On the other hand, if we _do_ queue the incremental updates anyway, then they'll repeat a bit of the same work, but it _should be_ harmless.

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/search/service/solr/indexer/AbstractSolrIndexUpdateServiceImpl.java
```diff
@@ -0,0 +1,292 @@
...
+    
+    @Override
+    public SolrInputDocument buildDocument(Indexable indexable) {
+        return commandHandler.buildDocument(indexable);
+    }
+    
+    @Override
+    public SolrInputDocument buildDocument(Indexable indexable, List<IndexField> fields, List<Locale> locales) {
+        return commandHandler.buildDocument(indexable, fields, locales);
+    }
```
ckittrell, line 268: It seems like these `buildDocument(...)` methods don't belong in this class - I'd expect to only see them in the CommandHandler. @ktisdell do you have a particular use case in mind for these methods?

ktisdell: @ckittrell I originally added them so that services outside of this class can build documents as needed.  Consider, for example, `WorkflowSolrIndexServiceImpl#getSolrDocuments(...)`.  In fact, this should likely be updated to optionally use this class for building documents rather than using the legacy class.

ktisdell: Although, they could likely just use the `CatalogDocumentBuilder` as well.

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/search/service/solr/indexer/AbstractSolrIndexUpdateCommandHandlerImpl.java
```diff
@@ -0,0 +1,271 @@
...
+            }
+            
+            if (command.getSolrInputDocuments() != null && ! command.getSolrInputDocuments().isEmpty()) {
+                addDocuments(collectionName, command.getSolrInputDocuments());
+                changeMade = true;
+            }
+            
+            try {
+                if (changeMade) {
+                    commit(collectionName, true, true, false);
```
ckittrell, line 94: While I agree that committing all deletes & additions at once is the right idea, there is potential for a huge commit if the command holds a large number of deletions and/or additions. 

Splitting into sets of smaller commits is likely the wrong idea since rollback scenarios get ugly quickly.

The best answer may be to enforce a limit on the number of deletions & additions per command. From there, it would be the responsibility of the caller to provide smaller commands.

Of course, this is assuming that there's a relatively low threshold for the max number of items that can be committed before seeing a significant performance degradation.

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/search/service/solr/indexer/AbstractSolrIndexUpdateServiceImpl.java
ktisdell: @ckittrell I don't disagree and don't have a strong opinion here.  I think more work could be done to have a coordinated foreground command queue and background command queue which could effectively handle situations like that.  Given that we don't, I'm OK with removing that logic in favor of "the extra work" that will have to be done by the indexer, assuming that doing that extra work simply restores the state after it is complete (i.e. causes no harm).

ckittrell: Agreed - lets hit the easy button here & remove the `!commandQueue.contains(FullReindexCommand.DEFAULT_INSTANCE)` portion of the condition.

ckittrell (approved):

PR #2238, opened 2020-01-21
Create a new list for email attachments on clone

marieStandeven:
**A Brief Overview**
Adding email attachments caused other email info objects created via the clone() method to receive those attachments incorrectly.

https://github.com/BroadleafCommerce/QA/issues/3908

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/email/service/info/EmailInfo.java
+++ b/common/src/main/java/org/broadleafcommerce/common/email/service/info/EmailInfo.java
@@ -152,7 +152,7 @@ public class EmailInfo implements Serializable {
 
     public synchronized EmailInfo clone() {
         EmailInfo info = new EmailInfo();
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/broadleafcommerce/common/email/service/info/EmailInfo.java
```diff
@@ -152,7 +152,7 @@ public void setEncoding(String encoding) {
 
     public synchronized EmailInfo clone() {
         EmailInfo info = new EmailInfo();
-        info.setAttachments(attachments);
+        info.setAttachments(new ArrayList<Attachment>());
```
danielcolgrove: Let's create a new ArrayList and copy the attachments.  If the "client" needs to alter what is in the attachments, that would be logic they would need to implement.  A clone should be a full copy.

PR #2242, opened 2020-02-05
Skus added to products are propagated to sandbox but cannot be approved.

oleksiimiroshnyk:
BroadleafCommerce/QA#3900
Collection propagation is doesn't create change details for alt sanbox flow

The problem that if we add only sku there is only change detail for sku. 
So, when propagation logic comes into play it unable to find change detail that corresponds to workflow item, and so workflow item is not clonned.

When I change default sku field it presists product and not sku and in this way it create change details for product & sku and so workflow item is clonned during propagation.

So I save product before saving additional sku, in this way I have all workflow/change details and it goes smooth. 
Probably there should be another way to create change details, but I didin't find the place where I can hook and create additional change details for product when change details for sku are created and handle only the case when it happens during collection.add.
Guess this issue happens to other collections as well, so maybe @jefffischer can suggest a way where change details for product/"collection holder entity" should be created ? Maybe use somehow related_to_table information ?
Maybe add in controller direct call to add change details for "collection holder entity" and link to workflow item ?

### DIFF
--- a/admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/server/service/handler/SkuCustomPersistenceHandler.java
+++ b/admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/server/service/handler/SkuCustomPersistenceHandler.java
@@ -700,6 +700,7 @@ public class SkuCustomPersistenceHandler extends CustomPersistenceHandlerAdapter
                 return entity;
             }
 
+            dynamicEntityDao.persist(adminInstance.getProduct());
             //persist the newly-created Sku
             adminInstance = dynamicEntityDao.persist(adminInstance);
 


jefffischer (changes requested):
I'd rather not change the behavior of the system to force WFSBI instances and clones where there were none before in the previous strategy. I'm talking about the requirement with this change that a Product clone is made for a change to a associated sandboxable collection. This would not have been the case previously.

jefffischer:
WorkflowEntityPersisterUpdateInterceptorExtensionHandler is responsible for distinguishing the ChangeDetails required for a user submitted change in the admin. This ends up getting placed in the ChangeDetailContext which informs the final persistence of ChangeDetail instances downstream.

oleksiimiroshnyk:
>> I'm talking about the requirement with this change that a Product clone is made for a change to a associated sandboxable collection. This would not have been the case previously.

This is why I'm saving product -> clone is created. There might be better solution to create a product clone ?

>>WorkflowEntityPersisterUpdateInterceptorExtensionHandler is responsible for distinguishing the ChangeDetails required for a user submitted change in the admin. This ends up getting placed in the ChangeDetailContext which informs the final persistence of ChangeDetail instances downstream.

Not sure I'm following how this can help.
if I put breakpoint in 
com.broadleafcommerce.enterprise.workflow.persistence.persister.WorkflowEntityPersisterUpdateInterceptorExtensionHandler#processInsert

and call this.setupDetails(id, fields, object, session); with parameters for product and after that let program run it will create change details for product & sku and everything works fine, but the issue is that I don't know that this is save for collection, entity and I need to find "container entity" & create change details for it.  I can't find anything that will help me to understand that and also I don't see the generic way to find "collection container entity" so I will not do if-else for all our entities that can have collections

@jefffischer looks like I can't understand the correct way you are talking about, maybe a quick call ?

jefffischer:
I don’t have a solution in mind. However, I was pointing out where change details are initiated, since you asked for that. 

On ChangeDetail, there is a “relatedEntity” field (or something like that). What is that field being set to in this use case?

oleksiimiroshnyk:
for product change details:
Derived_table=blc_product
related_to_table=blc_product

for sku
derived_table=blc_sku
related_to_table=blc_product

but related_to_table field is set in SimpleChangeDAO that is invoked later WorkflowEntityPersisterUpdateInterceptorExtensionHandler. Information for this field comes from description of workflow item

jefffischer:
Assign the issue to me. I’ll research it when I get a chance.

PR #2247, opened 2020-02-13
fix fields disabled once Approve

putrenkov:
**A Brief Overview**
After press Approve button not all fields was disabled in Content Item

**Link to QA issue**
QA-3781 - https://github.com/BroadleafCommerce/QA/issues/3781


epenning (changes requested):
Mostly looks good! I was testing this out, and looks like it also needs handling for `.move-scitem-down` in the same way that `.move-scitem-up` is handled. Once that's done, this should be all good.

putrenkov:
@epenning ye you are right, thanks)

epenning (approved):

PR #2248, opened 2020-02-13
Ability to remove bundle

oleksiimiroshnyk:
BroadleafCommerce/QA#3869
Ability to remove bundle

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceBundlingService.java
+++ b/common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceBundlingService.java
@@ -92,4 +92,12 @@ public interface ResourceBundlingService {
      */
     boolean checkForRegisteredBundleFile(String versionedBundleName);
 
+    /**
+     * returns bundle name that contain that file
+     * @param fileName file to find in the bungle
+     * @return first bundle name that has file, null otherwise
+     */
+    public List<String> findBundlesNameByResourceFileName(String fileName);
+
+    boolean removeBundle(String bundleName);
 }
--- a/common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceBundlingService.java
+++ b/common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceBundlingService.java
@@ -97,7 +97,7 @@ public interface ResourceBundlingService {
      * @param fileName file to find in the bungle
      * @return first bundle name that has file, null otherwise
      */
-    public List<String> findBundlesNameByResourceFileName(String fileName);
+    List<String> findBundlesNameByResourceFileName(String fileName);
 
     boolean removeBundle(String bundleName);
 }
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceBundlingService.java
```diff
@@ -92,4 +92,12 @@
      */
     boolean checkForRegisteredBundleFile(String versionedBundleName);
 
+    /**
+     * returns bundle name that contain that file
```
epenning:
```suggestion
     * Returns names of bundles which contain the given file
```

common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceBundlingService.java
```diff
@@ -92,4 +92,12 @@
...
+     * @param fileName file to find in the bungle
+     * @return first bundle name that has file, null otherwise
+     */
+    public List<String> findBundlesNameByResourceFileName(String fileName);
```
epenning:
```suggestion
    List<String> findBundlesNameByResourceFileName(String fileName);
```

`public` would be unnecessary here, since it's an interface

common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceBundlingServiceImpl.java
```diff
@@ -448,4 +450,30 @@ protected Charset getBundleCharSet() {
...
     }
+
+    @Override
+    public List<String> findBundlesNameByResourceFileName(String fileName){
+        List<String> result = new ArrayList<>();
+        for (Entry<String, BundledResourceInfo> bundleInfo : createdBundles.entrySet()) {
+            BundledResourceInfo value = bundleInfo.getValue();
+            for (String s : value.getBundledFilePaths()) {
+                String f = FilenameUtils.removeExtension(fileName);
+                if(s.startsWith(f)){
```
epenning, line 461: Would there be any possible conflict due to using `startsWith`?

I could imagine there being two files like `nav.css` and `nav-mobile.css`, in which case a change to `nav.css` would also find the bundle for `nav-mobile.css`. Although it's not likely to be a problem right now, since this is only used for invalidating the bundles, and files like this would likely exist in the same bundle anyway.

epenning (changes requested):
Looks good and pretty straightforward, I just had a couple of suggestions and one question

common/src/main/java/org/broadleafcommerce/common/resource/service/ResourceBundlingServiceImpl.java
oleksiimiroshnyk: @epenning thx for pointing that out but I think I did that for some good reason, I don't remember exactly but I believe that there was some suffix of some hash, so I can't compare names exactly

Vitalij-Voronkoff: @epenning I checked it to see why this check did. That make sense, because to file path hash added. For example it's looks like : nav-css-tf167635363.css

PR #2249, opened 2020-02-14
Add a new request mapping for readonly collection with tabs 

putrenkov:
**Link to QA issue**
QA-3923 - https://github.com/BroadleafCommerce/QA/issues/3923

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/controller/entity/AdminBasicEntityController.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/controller/entity/AdminBasicEntityController.java
@@ -814,6 +814,22 @@ public class AdminBasicEntityController extends AdminAbstractController {
     }
 
 
+    @RequestMapping(
+            value = "/{id}/{collectionField:.*}/{collectionItemId}/view/{tab:[0-9]+}/{tabName}",
+            method = RequestMethod.POST
+    )
+    public String viewReadOnlyCollectionItemTab(HttpServletRequest request, HttpServletResponse response, Model model,
+                                        @PathVariable  Map<String, String> pathVars,
+                                        @PathVariable(value="id") String id,
+                                        @PathVariable(value="collectionField") String collectionField,
+                                        @PathVariable(value="collectionItemId") String collectionItemId,
+                                        @PathVariable(value="tabName") String tabName,
+                                        @ModelAttribute(value = "entityForm") EntityForm entityForm) throws Exception {
+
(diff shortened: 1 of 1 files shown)


admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/controller/entity/AdminBasicEntityController.java
```diff
@@ -814,6 +814,22 @@ public String viewCollectionItemTab(HttpServletRequest request, HttpServletRespo
...
+    public String viewReadOnlyCollectionItemTab(HttpServletRequest request, HttpServletResponse response, Model model,
+                                        @PathVariable  Map<String, String> pathVars,
+                                        @PathVariable(value="id") String id,
+                                        @PathVariable(value="collectionField") String collectionField,
+                                        @PathVariable(value="collectionItemId") String collectionItemId,
+                                        @PathVariable(value="tabName") String tabName,
+                                        @ModelAttribute(value = "entityForm") EntityForm entityForm) throws Exception {
+
+        return showViewUpdateCollection(request, model, pathVars, id, collectionField, collectionItemId, ModalHeaderType.VIEW_COLLECTION_ITEM.getType(), entityForm, null);
+    }
```
epenning, line 830: It looks like this is currently identical to the contents of `viewCollectionItemTab` - should there be anything additional here to make it read-only?

The original issue requirements aren't super clear so I'm not fully sure of what's intended, but I suppose maybe there should be something around making the collection item read-only here? Like what's done in `showViewCollectionItem`:

```
// Since this is a read-only view, actions don't make sense in this context
EntityForm ef = (EntityForm) model.asMap().get("entityForm");
addAuditableDisplayFields(ef);
ef.removeAllActions();
ef.setReadOnly();
```

epenning (changes requested):

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/controller/entity/AdminBasicEntityController.java
putrenkov: @epenning I think the problem was in missing mapping with "view"

Add another request mapping for this URL
"/{id}/{collectionField:.*}/{collectionItemId}/view/{tab:[0-9]+}/{tabName}"

putrenkov (commented):
@epenning I think the problem was in missing mapping with "view"

Add another request mapping for this URL
"/{id}/{collectionField:.*}/{collectionItemId}/view/{tab:[0-9]+}/{tabName}"

PR #2250, opened 2020-02-17
Added check for read only table still allows reorder edits

putrenkov:
**A Brief Overview**
QA-3183 - https://github.com/BroadleafCommerce/QA/issues/3183


epenning (approved):
LGTM, solves the original QA issue!

PR #2271, opened 2020-03-11
Additional lazy fetch tuning

putrenkov:
**Link to QA issue**
QA-3961 - https://github.com/BroadleafCommerce/QA/issues/3961


danielcolgrove:
@putrenkov I was afraid of possible "session closed" issues.  We're going to need to do some testing with these changes and for sure run the integration tests locally.  If testing looks good but the integration test fails, need to try to understand why.

PR #2289, opened 2020-03-30
Fixed issue with order "between date" filter when it does not correctly encode a pipe delimiter in the URL

Vitalij-Voronkoff:
BroadleafCommerce/QA#3975


danielcolgrove:
This didn't work for me.  I'm not sure what you are attempting to do.  All we need is to encode the URL.  A pipe character "|" is an illegal character in a URL.  It needs to be replaced with "%7C".  If I change the url manually it works.  Somewhere in the JS there is a process that is using "|".  Even when I manually change it to "%7C" while the search works perfectly, there is still something in the JS that rebuilds the URL with "|".

Vitalij-Voronkoff:
@danielcolgrove reworked.

PR #2294, opened 2020-04-09
Added new field to ListGrid for storing css classes

Vitalij-Voronkoff:
BroadleafCommerce/QA#3982

**A Brief Overview**
Added new field to ListGrid for storing css classes.


danielcolgrove:
--Noteworthy--
#### ListGrid CssClasses
A new property has been added to the ListGrid Form Component - cssClasses.  Any additional css class names can now be added to this new property.  This was added to distinguish between classes (which usually represents the underlying Java class) and the cssClasses (which represent the style of the listgrid).  The Java classname is still used as part of the css class string:
```html
th:classappend="${listGrid.className + ' ' + listGrid.getCssClassNames()}"
```

PR #2295, opened 2020-04-13
Fixed an issue where Inventory check doesn't work properly for products with product options

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/4000


StanislavFedorov:
Looks like Jenkins build error unrelated with this commit itself

PR #2311, opened 2020-04-30
Fix adding Id filter for the any XREF ListGrid

putrenkov:
**A Brief Overview**
Fix NPE when adding Id Filter for the any XREF ListGrid (ex. Products list under category).

**Additional context**
https://github.com/BroadleafCommerce/QA/issues/4028


cade-rea:
It turns out that the framework already has support for this. Using a field id like "[fieldName]Target" enables filtering by the adorned target entity id.  
`com.broadleafcommerce.enterprise.workflow.admin.service.module.EnterpriseAdornedTargetListPersistenceModule#getAdornedTargetFilterMappings`
Line 177:
`.withFilterValues(cto.get(adornedTargetList.getCollectionFieldName() + "Target").getFilterValues())`

PR #2319, opened 2020-05-18
Fix js scripts that break ECMAScript5 compatibility

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/4043

@danielcolgrove 
I have tested this and it worked for me but i have some concerns.
To meet ECMAScript 5 requirements we in fact downgrade JS code of external libs from ECMAScript 6 to ECMAScript 5.

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/lib/plugins/jquery.mCustomScrollbar.js
+++ b/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/lib/plugins/jquery.mCustomScrollbar.js
@@ -366,6 +366,19 @@ plugin home: http://manos.malihu.gr/jquery-custom-content-scroller
 					$this.mCustomScrollbar("scrollTo",(mCSB_dragger.position().top-(mCSB_draggerDragY))+y,{moveDragger:true,trigger:"internal"});
 				}
 			}
+			function PixelsScrollTo(to){
+				if(!mCSB_dragger.data("preventAction")){
+					mCSB_dragger.data("preventAction",true);
+					$this.mCustomScrollbar("scrollTo",to,{trigger:"internal"});
+				}
+			}
+			function ScrollButtonsSpeed(){
+				var speed=$this.data("scrollButtons_scrollSpeed");
+				if($this.data("scrollButtons_scrollSpeed")==="auto"){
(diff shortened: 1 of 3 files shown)


admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/lib/plugins/jquery.mCustomScrollbar.js
```diff
@@ -366,6 +366,19 @@ plugin home: http://manos.malihu.gr/jquery-custom-content-scroller
 					$this.mCustomScrollbar("scrollTo",(mCSB_dragger.position().top-(mCSB_draggerDragY))+y,{moveDragger:true,trigger:"internal"});
 				}
 			}
+			function PixelsScrollTo(to){
```
cja769, line 369: Since this is a plugin it's good to add some block comments around these changes to indicate that we changed it and why so that when the plugin gets updated we'll know that we'll likely need to add these changes again

admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/vendor/jquery.fluidbox.js
```diff
@@ -144,7 +139,12 @@
 						} else {
 							data.imgScale = $w.width()*settings.viewportFill/$img.width();
 						}						
-					}
+					};
+
+					var $img	= $fbItem.find('img'),
```
cja769, line 144: Same as the other comment. Just add some block comments indicating that this was changed and why it was changed

cja769 (changes requested):
@StanislavFedorov I checked to see if ECMASCRIPT_2016 worked with current private demo and it was breaking in the same ways as it does with your branch so no worries on that front. We can open up another ticket to address that but ECMASCRIPT5 does seem to work with this branch. Along with the requested changes I would also like you to do `git rebase -i`, do `fixup` on the commit that adds these changes, and then `reword` the original commit. The new commit should explain why these changes were made (i.e. why these changes actually fixed the problem). Since these plugins very rarely change we need to have good records on what we've done to them or we have to do this all over again the next time the plugins are updated. I didn't heavily test this but the normal flows seem to have worked and I didn't get any errors in minification.

PR #2331, opened 2020-06-02
Offer rules should allow between rules

oleksiimiroshnyk:
BroadleafCommerce/QA#4061
Offer rules should allow between rules

Changed logic so checkForInvalidSubGroup will check only dtos created from subgroup

Changed test because it should be a valid case according to Brian & Daniel

### DIFF
--- a/admin/broadleaf-admin-module/src/test/java/org/broadleafcommerce/admin/web/rulebuilder/MVELToDataWrapperTranslatorTest.java
+++ b/admin/broadleaf-admin-module/src/test/java/org/broadleafcommerce/admin/web/rulebuilder/MVELToDataWrapperTranslatorTest.java
@@ -301,7 +301,22 @@ public class MVELToDataWrapperTranslatorTest extends TestCase {
         entities[0] = entity;
 
         DataWrapper dataWrapper = translator.createRuleData(entities, "matchRule", null, null, fulfillmentGroupFieldService);
-        assert(dataWrapper.getError().equals(MVELToDataWrapperTranslator.SUB_GROUP_MESSAGE));
+        assert(dataWrapper.getData().size() == 1);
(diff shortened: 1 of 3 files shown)


admin/broadleaf-admin-module/src/test/java/org/broadleafcommerce/admin/web/rulebuilder/MVELToDataWrapperTranslatorTest.java
```diff
@@ -301,7 +301,22 @@ public void testNestedExpressionExceptionForFulfillmentGroupQualificationDataWra
         entities[0] = entity;
 
         DataWrapper dataWrapper = translator.createRuleData(entities, "matchRule", null, null, fulfillmentGroupFieldService);
-        assert(dataWrapper.getError().equals(MVELToDataWrapperTranslator.SUB_GROUP_MESSAGE));
+        assert(dataWrapper.getData().size() == 1);
```
ckittrell, line 304: This test should be renamed since it no longer causes an error

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/rulebuilder/dto/DataDTO.java
```diff
@@ -35,6 +35,7 @@
     protected Long previousContainedPk;
     protected Integer quantity;
     protected String condition;
+    protected boolean createdFromSubGroup;
```
ckittrell: This should be explicitly defaulted to `false`

ckittrell (approved):
Overall this looks good, but there are a couple of minor details that are worth cleaning up before merging this PR.

PR #2333, opened 2020-06-09
Fix issue where demo SQL import statements that had escaped characters and newlines didn't import correctly

cja769:
Previously we prepended an "E" to the beginning of all string values because it makes postgres read escaping correctly. However additional code was added afterwards to replace new lines with the postgres special characters for new lines which broke the string values into multiple string values. Consequently this caused every block of strings after the first to no longer have the "E" in front of it therefore the escapes were no longer being evaluated. This mainly affected the content item for the homepage which is HTML that had escaped double quotes and new line characters.


cja769:
This change is being made in 5.0.x because we need to be able to test things against postgres with 5.0 and without this change it makes it very annoying.

cja769:
https://github.com/BroadleafCommerce/QA/issues/4071

PR #2343, opened 2020-06-12
Moving the profile entities from the blStandardElements to a new blCountryElements cache region

danielcolgrove:


core/broadleaf-profile/src/main/java/org/broadleafcommerce/profile/core/domain/CustomerAttributeImpl.java
```diff
@@ -40,7 +40,7 @@
 @Entity
 @Inheritance(strategy = InheritanceType.JOINED)
 @Table(name="BLC_CUSTOMER_ATTRIBUTE")
-@Cache(usage = CacheConcurrencyStrategy.READ_WRITE, region="blStandardElements")
+@Cache(usage = CacheConcurrencyStrategy.READ_WRITE, region="blProfileElements")
```
cja769: Feels like this should be in `blCustomerElements` honestly

core/broadleaf-profile/src/main/java/org/broadleafcommerce/profile/core/domain/ChallengeQuestionImpl.java
```diff
@@ -36,7 +36,7 @@
 @Entity
 @Inheritance(strategy = InheritanceType.JOINED)
 @Table(name = "BLC_CHALLENGE_QUESTION")
-@Cache(usage = CacheConcurrencyStrategy.READ_WRITE, region="blStandardElements")
+@Cache(usage = CacheConcurrencyStrategy.READ_WRITE, region="blProfileElements")
```
cja769: This can probably just stay in `blStandardElements` it's hardly ever used and if we are going to change `blProfileElements` to `blCountryElements` or whatever then it doesn't really fit there.

cja769 (commented):
Seems like the region should be called `blCountryElements` or something along those lines. Seems more concise to put all of the country domain stuff in its own region so that we can just put all of those tables in cache but if it's `blProfileElements` it has a larger umbrella term.

cja769 (approved):

PR #2344, opened 2020-06-12
Added new purge process that is based on a purge specification.

danielcolgrove:


cja769 (changes requested):
Since we do have control over how the transactions/connections/statements are handled I would like to be aware of https://stackoverflow.com/a/37651864/4273701. It seems that a transaction is started implicitly even on read operations with postgres so we'll want to manage those transactions properly and make sure they get committed or rolledback

danielcolgrove:
It's handled in the ResourcePurgeServiceImpl.  Same place that it was handled in the legacy purge.  Do you think that should be redesigned for this change?

cja769:
Yep I see that now. Sorry about that. Good to merge.

cja769 (approved):

PR #2347, opened 2020-06-16
Upgrade lib

putrenkov:
**A Brief Overview**
Libs upgrade:
Spring Security: 5.0.12 -> 5.0.17

**Additional context**
BroadleafCommerce/QA#4080


riteshadhikari17:
@danielcolgrove currently its 5.0.x but this is bumping up to 5.1.x minor release. Is it ok? 5.0.16 has the fix and 5.0.17 is the latest.

danielcolgrove:
I would feel more comfortable going to the latest of the current version and not jumping to a new release version.

PR #2352, opened 2020-06-25
Add default value to Content Items

putrenkov:
**A Brief Overview**
Broadleaf does not have support for default values for Content Items.
Added a new colum defaultValue to table BLC_FLD_DEF
and if we add new field it will set to defaultValue if it have such.

BroadleafCommerce/QA#4031


cja769:
This looks good however since it makes a domain change it'll have to wait until 6.2 which we haven't really discussed much so this have to be left open for a while. Please rebase this on top of 6.1 (you'll have to update the ehcache stuff to jcache)

cja769:
Closing in favor of https://github.com/BroadleafCommerce/BroadleafCommerce/pull/2353

PR #2355, opened 2020-07-01
Fix issue where a new customer segment could not be added from the offer page

oleksiimiroshnyk:
BroadleafCommerce/QA#4094
Fix add customer segment on offer page


putrenkov:
@oleksiimiroshnyk @danielcolgrove I tested looks good for me

PR #2356, opened 2020-07-02
Cannot add additional content data

putrenkov:
**A Brief Overview**
If got same names of field group with master field group and not master field group,
it will removed additional content data. Solution is make name unique to avoid unwanted remove

**Additional context**
BroadleafCommerce/QA#4035


oleksiimiroshnyk:
probably need to add to change notes that we add unique constraint to field group name
@riteshadhikari17 FYI ^^

PR #2358, opened 2020-07-08
Historical Purge

oleksiimiroshnyk:
BroadleafCommerce/QA#3926
Historical Purge initial commmit

For now only class the builds sql.
Need to create or describe job creation that will trigger class that builds sql.
Need to have some conditions when job is triggered
Need to have some props to enable/disable logic
Think about archiving order first ?
Need to solr reindex orders? or delete from solr deleted orders ?


oleksiimiroshnyk:
this is for 5.2, but should be 6

PR #2362, opened 2020-07-13
Remove usage deprecated methods

putrenkov:
**A Brief Overview**
Refactor and remove deprecated method

**Link to QA issue**
BroadleafCommerce/QA#3878

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/security/handler/AdminUserCustomPersistenceHandler.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/security/handler/AdminUserCustomPersistenceHandler.java
@@ -24,7 +24,6 @@ import org.apache.commons.logging.LogFactory;
 import org.broadleafcommerce.common.exception.ServiceException;
 import org.broadleafcommerce.common.persistence.Status;
 import org.broadleafcommerce.common.presentation.client.OperationType;
-import org.broadleafcommerce.common.util.BLCSystemProperty;
 import org.broadleafcommerce.openadmin.dto.Entity;
 import org.broadleafcommerce.openadmin.dto.FieldMetadata;
 import org.broadleafcommerce.openadmin.dto.PersistencePackage;
@@ -38,6 +37,8 @@ import org.broadleafcommerce.openadmin.server.security.service.AdminSecurityServ
 import org.broadleafcommerce.openadmin.server.service.ValidationException;
 import org.broadleafcommerce.openadmin.server.service.handler.CustomPersistenceHandlerAdapter;
 import org.broadleafcommerce.openadmin.server.service.persistence.module.RecordHelper;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.core.env.Environment;
 import org.springframework.stereotype.Component;
 
 import java.util.Map;
@@ -57,9 +58,12 @@ public class AdminUserCustomPersistenceHandler extends CustomPersistenceHandlerA
 
     @Resource(name = "blAdminSecurityRemoteService")
     protected SecurityVerifier adminRemoteSecurityService;
+    
+    @Autowired
+    protected Environment environment;
 
     protected boolean getRequireUniqueEmailAddress() {
-        return BLCSystemProperty.resolveBooleanSystemProperty("admin.user.requireUniqueEmailAddress");
+        return  environment.getProperty("admin.user.requireUniqueEmailAddress").isEmpty();
     }
 
     @Override
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/security/handler/AdminUserCustomPersistenceHandler.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/security/handler/AdminUserCustomPersistenceHandler.java
@@ -37,7 +37,6 @@ import org.broadleafcommerce.openadmin.server.security.service.AdminSecurityServ
 import org.broadleafcommerce.openadmin.server.service.ValidationException;
 import org.broadleafcommerce.openadmin.server.service.handler.CustomPersistenceHandlerAdapter;
 import org.broadleafcommerce.openadmin.server.service.persistence.module.RecordHelper;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.core.env.Environment;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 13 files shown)


admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/security/handler/AdminUserCustomPersistenceHandler.java
```diff
@@ -57,9 +58,12 @@
 
     @Resource(name = "blAdminSecurityRemoteService")
     protected SecurityVerifier adminRemoteSecurityService;
+    
+    @Autowired
+    protected Environment environment;
 
     protected boolean getRequireUniqueEmailAddress() {
-        return BLCSystemProperty.resolveBooleanSystemProperty("admin.user.requireUniqueEmailAddress");
+        return  environment.getProperty("admin.user.requireUniqueEmailAddress").isEmpty();
```
cja769: This should not have a `isEmpty`. That is a boolean property. Should be 
```
return environment.getProperty("admin.user.requireUniqueEmailAddress", Boolean.class);
```

core/broadleaf-framework-web/src/main/java/org/broadleafcommerce/core/web/controller/account/BroadleafManageWishlistController.java
```diff
@@ -48,7 +49,7 @@
     protected static String accountWishlistRedirect = "redirect:/account/wishlist";
 
     public String add(HttpServletRequest request, HttpServletResponse response, Model model,
-            AddToCartItem itemRequest, String wishlistName) throws IOException, AddToCartException, PricingException  {
+                      ConfigurableOrderItemRequest itemRequest, String wishlistName) throws IOException, AddToCartException, PricingException  {
```
cja769: This needs to be OrderItemRequestDTO instead. First of all for compatibility reasons (i.e. clients may be using this method and sending it AddToCartItem which subclasses OrderItemRequestDTO however it doesn't subclass ConfiguratbleOrderItemRequest) and second of all if we only need it to be an OrderItemRequestDTO then there's not reason to require a more specific implementation

core/broadleaf-framework-web/src/main/java/org/broadleafcommerce/core/web/controller/cart/BroadleafCartController.java
```diff
@@ -350,7 +350,7 @@ public String remove(HttpServletRequest request, HttpServletResponse response, M
 
     @Deprecated
     public String remove(HttpServletRequest request, HttpServletResponse response, Model model,
-                         AddToCartItem itemRequest) throws IOException, PricingException, RemoveFromCartException {
+                         ConfigurableOrderItemRequest itemRequest) throws IOException, PricingException, RemoveFromCartException {
```
cja769: This needs to be OrderItemRequestDTO instead. First of all for compatibility reasons (i.e. clients may be using this method and sending it AddToCartItem which subclasses OrderItemRequestDTO however it doesn't subclass ConfiguratbleOrderItemRequest) and second of all if we only need it to be an OrderItemRequestDTO then there's not reason to require a more specific implementation

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/call/AddToCartItems.java
```diff
@@ -28,32 +28,32 @@
     @SuppressWarnings("unchecked")
 
     //TOOD: this should probably be refactored to be called "rows" like in other model objects
-    private List<AddToCartItem> addToCartItems =   LazyList.decorate(
-            new ArrayList<AddToCartItem>(),
-            FactoryUtils.instantiateFactory(AddToCartItem.class));
+    private List<ConfigurableOrderItemRequest> addToCartItems =   LazyList.decorate(
+            new ArrayList<ConfigurableOrderItemRequest>(),
+            FactoryUtils.instantiateFactory(ConfigurableOrderItemRequest.class));
```
cja769: These should be OrderItemRequestDTO instead. No need to make it more specific when it doesn't need to be

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/call/AddToCartItems.java
```diff
@@ -28,32 +28,32 @@
...
 
     private long productId;
     private long categoryId;
 
     public void setProductId(long productId) {
         this.productId = productId;
-        for(AddToCartItem addToCartItem : addToCartItems) {
+        for(ConfigurableOrderItemRequest addToCartItem : addToCartItems) {
```
cja769: These should be OrderItemRequestDTO instead. No need to make it more specific when it doesn't need to be

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/call/AddToCartItems.java
```diff
@@ -28,32 +28,32 @@
...
             addToCartItem.setProductId(productId);
         }
     }
 
     public void setCategoryId(long categoryId) {
         this.categoryId = categoryId;
-        for(AddToCartItem addToCartItem : addToCartItems) {
+        for(ConfigurableOrderItemRequest addToCartItem : addToCartItems) {
```
cja769: These should be OrderItemRequestDTO instead. No need to make it more specific when it doesn't need to be

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/call/AddToCartItems.java
```diff
@@ -28,32 +28,32 @@
...
             addToCartItem.setCategoryId(categoryId);
         }
     }
 
-    public List<AddToCartItem> getAddToCartItems() {
+    public List<ConfigurableOrderItemRequest> getAddToCartItems() {
```
cja769: These should be OrderItemRequestDTO instead. No need to make it more specific when it doesn't need to be

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/call/AddToCartItems.java
```diff
@@ -28,32 +28,32 @@
...
         return addToCartItems;
     }
 
-    public void setAddToCartItem(List<AddToCartItem> addToCartItems) {
+    public void setAddToCartItem(List<ConfigurableOrderItemRequest> addToCartItems) {
```
cja769: These should be OrderItemRequestDTO instead. No need to make it more specific when it doesn't need to be

cja769 (changes requested):
I was afraid that these changes may need to be added into 6.2 however it looks like the changes made shouldn't affect clients. With the requested changes it should be good to merge

core/broadleaf-framework-web/src/main/java/org/broadleafcommerce/core/web/controller/account/AbstractCustomerAddressController.java
```diff
@@ -87,8 +91,8 @@ protected void initBinder(HttpServletRequest request, ServletRequestDataBinder b
         initBinderService.configAddressInitBinder(binder);
     }
 
-    protected List<State> populateStates() {
-        return stateService.findStates();
+    protected List<CountrySubdivision> populateStates() {
+        return countrySubdivisionService.findSubdivisions();
```
cja769: Lets also revert this change. It looks like that in our demo we extend this and set a model attribute on it. Also CountrySubdivision != State therefore the front end mappings setup in demo site aren't setup correctly for this change (even in the PR). We'll catch this with the next pass when we do 6.2

putrenkov:
@cja769 I pushed my changes

PR #2366, opened 2020-07-14
Fixed MVEL internal quotes escaping

jetchegoyen:
BroadleafCommerce/QA#4108
Fixed MVEL internal quotes escaping

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/rulebuilder/DataDTOToMVELTranslator.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/rulebuilder/DataDTOToMVELTranslator.java
@@ -359,7 +359,7 @@ public class DataDTOToMVELTranslator {
 
     protected String escapeInternalQuotes(String processedValue) {
         String regex = "(?<!^)(?<!^\\s)\\\"(?!\\s$)(?!$)";
-        return processedValue.replaceAll(regex, "\\\\u0022");
+        return processedValue.replaceAll(regex, "\\\\\"");
     }
 
     public boolean isProjection(Object value) {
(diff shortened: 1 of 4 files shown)


danielcolgrove:
Setup a product with a quoted name and the rendering of the rule builder wasn't correct.  It showed the escaped value.  Product was

Insanity Sauce 8" "hi-top" bottle

Screen Shot 2020-07-14 at 7 23 02 PM

oleksiimiroshnyk:
there are couple of issues:
1. Reported by the client, MVEL doesn't allow unbalanced quotes. Like if you choose to wrap string elements of the "array" to be double quote - " then you have to have even number of those quotes, and doesn't matter escape you them or not.
So it will allow you - some"thing", but will not allow you some"thing, but will allow you some'thing because we use double quote to wrap string elements in array.

The only solution I found is to replace actual string elements to variables and assign string values to those variables.
For example:
Collection.intersect(orderItem.name,["blah-blah","bad\"string"]).size>0 -> will be changed to:
Collection.intersect(orderItem.name,[var1,var2]).size>0 , where var1=blah-blah, var2=bad\"string

2. Another issue is that when you have added some"thing" to rule builder and want to add more products, for example just Sudden Death after save, reload, save etc you will end up that value printed will be some"thing Sudden Death .
Closing quote will be missing, and after some save this corrupted string also will be persisted to DB.
This is because we have a regexp that is too aggressive for replacement.

oleksiimiroshnyk:
@jetchegoyen I pushed a fix.
See my prev. comment of the reasons I did such fix

this requires careful testing as some cases might be missing

oleksiimiroshnyk:
looks like if we decide to go with this solution, it should support quite a lot of different cases - now it fails on tests, because it doesn't expect [eval("aaa","bbb", eval("ddd,"ddd"))] contains eval("sss","yyy")

danielcolgrove:
Can you update the tests so they will pass?

oleksiimiroshnyk:
ok, seems I missed the working case:
expression:
CollectionUtils.intersection(orderItem.?name,[\"5\\u0022 wrench\"]).size()>0
works
EDIT: looks like formatting problems: you need 2 slashes before u0022

so my fix can be rolled back for the AbstractBaseController

fix for blc_admin.js probably needs to be revisited. and myabe that's the place BLCAdmin.unescapeString where we can replace u0022 with quotes ?

jetchegoyen:
just tested with **Insanity Sauce 8" "hi-top" bottle** and so far is working fine.
I updated `ruleBuilder.html` template replacing the escaped string, and looks good on the ui.

`data-data=${#strings.replace(field.dataWrapper.serialize(), '\u0022', '&quot;')},`

PR #2367, opened 2020-07-20
Fix getting correct catalog

putrenkov:
**A Brief Overview**
Fix getting correct catalog when standard site has multiple catalogs

**Link to QA issue**
BroadleafCommerce/QA#4110


cja769:
I don't understand why it's named after static assets. What's the reasoning behind the naming? Also this concept of setting the catalog based on operating on a catalog discriminated entity probably should be done deeper in the admin flow. More specifically, this probably needs to be done closer to persistence in some way. This is just one of likely many entry points to creating a relationship.

putrenkov:
Depends on BroadleafCommerce/BroadleafCommerce#358

putrenkov:
@cja769 I pushed changes, but not sure where could put trigger for my handler

cja769:
@oleksiimiroshnyk I tested these changes and it's hitting the extension manager and everything but the new piece in the BroadleafAdminRequestProcessor doesn't seem to be getting hit at all. I checked from the client side and I'm not seeing the parameter in the request.

oleksiimiroshnyk:
@cja769 are you sure you have everything up to date, rebuild etc ?
I'm able to put breakpoint in BroadleafAdminRequestProcessor and stop there. Also UI/html has changes..

As a side effect, now all status, ribbon etc calls are sent with that parameter.

xxx

oleksiimiroshnyk:
@cja769 found why you don't see it.
Sorry, forgot to add enterprise project fix

cja769 (approved):

PR #2372, opened 2020-07-27
Add Translation Page to Admin

michaelkloc:
**Overview**
Add section to Catalog Module
Add Maintain Permission to Master Admin
Add Controller for Translation Entity List


admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/web/controller/entity/AdminTranslationEntityController.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package org.broadleafcommerce.admin.web.controller.entity;
+
+import org.broadleafcommerce.openadmin.web.controller.entity.AdminBasicEntityController;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.RequestMapping;
+
+import java.util.Map;
+
+@Controller("blAdminProductController")
```
cja769, line 26: I don't believe this is the bean id you meant to use. When starting up this causes the following exception
```
Caused by: org.springframework.context.annotation.ConflictingBeanDefinitionException: Annotation-specified bean name 'blAdminProductController' for bean class [org.broadleafcommerce.admin.web.controller.entity.AdminTranslationEntityController] conflicts with existing, non-compatible bean definition of same name and class [org.broadleafcommerce.admin.web.controller.entity.AdminProductController]
```

cja769 (changes requested):

riteshadhikari17:
Closing this. https://github.com/BroadleafCommerce/QA/issues/4132 covers it.

PR #2375, opened 2020-07-30
Add support of sku name contains MVEL rule to Solr criteria translation

oleksiimiroshnyk:
BroadleafCommerce/QA#4121
Add support of sku name contains MVEL rule to Solr criteria translation
Error message chagne


oleksiimiroshnyk:
this can be probably closed in favor of BroadleafCommerce/QA#4129

PR #2377, opened 2020-07-31
Add Spring dev tools

putrenkov:
**A Brief Overview**
Add Spring dev tools

**Additional context**
BroadleafCommerce/QA#4118

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -424,8 +424,8 @@
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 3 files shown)


cja769 (commented):
Any reason this property is here and not in the demos? The Spring dev tools dependency is in the starter projects so I would expect this property to live there instead

cja769:
Closing since these changes should live in ReferenceSite and DemoSite

PR #2378, opened 2020-08-04
Track that offer code was removed from the cart because of expiration and notify user

oleksiimiroshnyk:
BroadleafCommerce/QA#4130
Track that offer code was removed from the cart because of expiration and notify user

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/offer/service/OfferServiceImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/offer/service/OfferServiceImpl.java
@@ -26,6 +26,7 @@ import org.broadleafcommerce.common.persistence.EntityDuplicator;
 import org.broadleafcommerce.common.sandbox.SandBoxHelper;
 import org.broadleafcommerce.common.util.StreamCapableTransactionalOperationAdapter;
 import org.broadleafcommerce.common.util.StreamingTransactionCapableUtil;
+import org.broadleafcommerce.common.web.BroadleafRequestContext;
 import org.broadleafcommerce.core.offer.dao.CustomerOfferDao;
 import org.broadleafcommerce.core.offer.dao.OfferCodeDao;
 import org.broadleafcommerce.core.offer.dao.OfferDao;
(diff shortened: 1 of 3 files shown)


core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/offer/service/OfferServiceImpl.java
```diff
@@ -198,7 +201,13 @@ public OfferCode lookupOfferCodeByCode(String code){
             }
         }
         List<OfferCode> orderOfferCodes = refreshOfferCodesIfApplicable(order);
+        int before = orderOfferCodes.size();
         orderOfferCodes = removeOutOfDateOfferCodes(orderOfferCodes);
+        int after = orderOfferCodes.size();
+        Object o = BroadleafRequestContext.getBroadleafRequestContext().getAdditionalProperties().get(FINALIZE_CHECKOUT);
+        if(o !=null && (Boolean)o && before!=after){
+            BroadleafRequestContext.getBroadleafRequestContext().getAdditionalProperties().put(OfferServiceImpl.OFFERS_EXPIRED, Boolean.TRUE);
+        }
```
cja769, line 210: Would we be able to move this logic into the actual checkout workflow? Being a service method I don't feel like its job is to be dealing with these kinds of things whereas the actual checkout workflow is actively managing the order and order state so it would make more sense for it to live there. That may require a new method to be made in the OfferService so that the checkout workflow can check to see if offers that were added expired before updating the offers on the order. Additionally, just looking at the code, it seems like this may only apply to offers added via an offer code. What about cases where offers were automatically added but expired before checkout?

cja769:
Sorry i thought I had submitted that review days ago but I guess I never actually submitted it. That's my mistake.

oleksiimiroshnyk:
@cja769 created another PR https://github.com/BroadleafCommerce/BroadleafCommerce/pull/2391/
Thx, I missed automatically considered offers.

when you say "actual checkout workflow" can you point to some classes ?
The issue I see that offer codes are removed in one place, offers in another place and error is shown in completely different place, so had to use "flags"
In this new PR I moved offer code removal to the place where expired offers are removed...should be a little better I think,
Anyway this is a legacy AFAIK new microservices will be this year, so it is investment in a refactoring of legacy...

PR #2384, opened 2020-08-16
Fixed WYSIWYG editor to allow adding columns and rows to an HTML table

danielcolgrove:
Fixes https://github.com/BroadleafCommerce/QA/issues/4141


cja769 (commented):
Should we add some comments indicating that we changed these lines for future reference? I'm not sure what the best way of modifying plugins is so that if/when this library gets upgraded we don't run into this issue again

PR #2391, opened 2020-08-20
Expired offer code refactoring

oleksiimiroshnyk:
Fixes BroadleafCommerce/QA#4130
Expired offer code refactoring


cja769:
Moving the conversation from the other PR to here
> when you say "actual checkout workflow" can you point to some classes ?
The issue I see that offer codes are removed in one place, offers in another place and error is shown in completely different place, so had to use "flags"
In this new PR I moved offer code removal to the place where expired offers are removed...should be a little better I think,
Anyway this is a legacy AFAIK new microservices will be this year, so it is investment in a refactoring of legacy...

`org.broadleafcommerce.core.pricing.service.workflow.OfferActivity` is the class that actually calls `OfferServiceImpl#applyAndSaveOffersToOrder`. My thought is that `applyAndSaveOffersToOrder`'s job is to simply do what it says which is to apply and save offers to the order. It doesn't and shouldn't have any logic in relation to the front end or anything like that. However, `OfferActivity` is one of many activities in the offer workflow which directly relates to a user action on a website. It's a thin difference but it's difference between service code (effectively code that has a very defined job) and what our workflows/activities do which inherently handles some of these caveats where you have to do weird stuff like this. Looking at your code it builds out perfectly to put the code in `OfferActivity` because you count the offers before calling `applyAndSaveOffersToOrder` and then count after you call it. There's an argument to be made that this logic should be done every time `applyAndSaveOffersToOrder` is called but, from what I can tell, it'd be cleaner to put that code in `OfferActivity` and leave `applyAndSaveOffersToOrder` to simply update the offers on an order and not worry about additional properties on the `BRC`. The actual logic looks good to me. Unfortunately this concept wasn't written in originally and this is really the best way to handle it as the code is built currently however much I hate using the `BRC` for storing random values.

oleksiimiroshnyk:
pushed a fix.
@cja769 I moved logic to OfferActivity, but I don't understand your comment:
"There's an argument to be made that this logic should be done every time applyAndSaveOffersToOrder is called "

As far as I see it will be called every time, maybe I'm missing something

cja769:
@oleksiimiroshnyk I was just making a point that where you made the change originally made sense because by putting it there then the logic to calculate if an offer was removed is ran ever time that method is called however I like the logic living in the activity more. This looks good to me by the way

cja769 (approved):

PR #2405, opened 2020-09-11
Use locale specific number formats to check numbers

oleksiimiroshnyk:
Use locale specific number formats to check numbers
Fixes: BroadleafCommerce/QA#4182


danielcolgrove:
@oleksiimiroshnyk I pulled down this branch locally to test and I can't create an offer.  I don't get any exceptions but the new entity modal is popping up which is an indication of an error.  The enterprise-add call returns a 302 but it should return a 200.

oleksiimiroshnyk:
@danielcolgrove this means some validation error occurred.
I tested again on France & Spain locales, and it works fine - I can see full screen entity form, able to save, promote etc.
Any more details so I can test locally ?

riteshadhikari17:
I didn't have issues. But are these expected?
Euro
image
Spanish(Mexico)
image

oleksiimiroshnyk:
@riteshadhikari17 yes. From the code point of view this is expected, because "default" from blc_currency is driving this.

Issue #2417, opened 2020-09-27
POM refactoring


pedsf1968:
- Branch : develop-6.0.x

Write all <dependencies> and plugins <versions> in <properties> section and use defined properties.
Sort <properties> in alphabetical order.

beautify POM sections to separate them
![image](https://user-images.githubusercontent.com/20228762/94360749-23374180-00b0-11eb-8255-8ed1aa39bcae.png)

![image](https://user-images.githubusercontent.com/20228762/94360759-39dd9880-00b0-11eb-91e5-b2d0726d09bb.png)

pedsf1968:
create feature centralise-string-const

PR #2420, opened 2020-10-03
POM refactoring

pedsf1968:
Refactoring all POM #2417

Declare all POMs childrens dependencies in POM parent.
Set all versions as properties in POM parent
Separate all sections

### DIFF
--- a/admin/broadleaf-admin-functional-tests/pom.xml
+++ b/admin/broadleaf-admin-functional-tests/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.broadleafcommerce</groupId>
         <artifactId>admin</artifactId>
-        <version>6.0.11-SNAPSHOT</version>
+        <version>6.0.12-SNAPSHOT</version>
     </parent>
     <artifactId>broadleaf-admin-functional-tests</artifactId>
     <name>BroadleafCommerce Admin Functional Tests</name>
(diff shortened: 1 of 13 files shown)


pedsf1968:
Why the check is not compliant?

danielcolgrove:
@pedsf1968 We appreciate your interest in Broadleaf.  While our licensing allows you to use our software for free (based on the terms of the license agreement (http://license.broadleafcommerce.org/fair_use_license-1.0.txt), it is difficult for us to accept changes and updates from community developers.  There are several reasons why, one of which is that we provide commercial licensing with terms that requires us to be very careful with what and how features get added to the framework.  Please know that community users can fork, extend, and customize as requirements demand (again, within the terms of the Broadleaf Fair Use License Agreement) and you are certainly welcome to use the software for your own needs.

danielcolgrove:
Regarding this change, we use a BOM approach for managing dependencies.  So the parent pom.xml doesn't have direct references to the individual dependencies.  Instead it references a "parent" artifact and version (broadleaf-boot-starter-parent).  That parent then references a BOM with all the dependencies.

See: https://github.com/BroadleafCommerce/DemoSite/blob/0b31154597635d1b3e3a9bdeed4041a1480cc4e8/pom.xml#L5

http://nexus.broadleafcommerce.org/nexus/content/groups/community-source-releases/org/broadleafcommerce/broadleaf-boot-starter-parent/6.1.3-GA/broadleaf-boot-starter-parent-6.1.3-GA.pom

and

http://nexus.broadleafcommerce.org/nexus/content/groups/community-source-releases/com/broadleafcommerce/broadleaf-bom/6.1.3-GA/broadleaf-bom-6.1.3-GA.pom

PR #2423, opened 2020-10-08
Added the ability to override the promotion calculation logic to allow an item percent off promotion to use more precise discounting

bpolster:
**Feature Description.**
This feature will allow clients to override the promotion calculation logic to allow an item percent off promotion to use more precise discounting.

For example, consider a product that is 77.85 and an offer that is 10% off.

By default, the system will use BankersRounding (HALF_EVEN) and round the adjustment amount.

So, for a quantity of 1 the out of box calculated price would be: 77.85 - 7.78 = $70.07 with a savings of 7.78
For a quantity of 3, the out of box calculated price would be: (77.85 * 3) - (7.78 * 3) = $210.21 with a savings of $23.34.

A more precise representation would be to store the adjustments and not round until after the multiplication.
So, (77.85 * .1) * 3 = 23.355 which would be rounded to 23.36 (using bankers rounding).

To support this use case, code has been added to support a client override of the default rounding mode.

Clients can control both the rounding mode and the precision.

**Steps to Implement**

Use the following properties to control the scale and rounding mode for percent off item offers.

// Integer value from java.lang.RoundingMode.  Leave blank/unset for default behavior
item.offer.percent.rounding.mode=DOWN

// Can be any value from 0 to 5, typical would be 3-5 (2 is the default for most currencies)
item.offer.percent.rounding.scale=5

**About the fix**
Changes the default behavior of PromotableItemFactoryImpl to impact the scale and rounding mode for the CandidateItemOffer. Other offer calculation functionality honors these scale settings.
Money utility updated to support multiply with explicit rounding.
OrderItemPriceDetail updated to read configured property when calculating total item adjustments.

Fixes https://github.com/BroadleafCommerce/QA/issues/4206

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/money/Money.java
+++ b/common/src/main/java/org/broadleafcommerce/common/money/Money.java
@@ -220,6 +220,14 @@ public class Money implements Serializable, Cloneable, Comparable<Money>, Extern
         return multiply(value);
     }
 
+    public Money multiplyWithRounding(int inInt, RoundingMode roundingMode) {
+        BigDecimal bd = getAmount();
+        BigDecimal in = new BigDecimal(inInt);
+        BigDecimal result = bd.multiply(in);
+        result = result.setScale(getCurrency().getDefaultFractionDigits(), roundingMode);
+        return new Money(result, getCurrency());
+    }
+
     public Money multiply(BigDecimal multiplier) {
         return new Money(amount.multiply(multiplier), currency, amount.scale() == 0 ? BankersRounding.getScaleForCurrency(currency) : amount.scale());
     }
@@ -470,6 +478,18 @@ public class Money implements Serializable, Cloneable, Comparable<Money>, Extern
         // Write out the client properties from the server representation.
         out.writeFloat(amount.floatValue());
         // out.writeObject(currency);
+    }
+
+    public static Money trimUnnecessaryScaleToCurrency(Money money) {
+        int currencyScale = 2;
+        if (money.getCurrency() != null) {
+            currencyScale = money.getCurrency().getDefaultFractionDigits();
+        }
+
+        BigDecimal amount = money.getAmount().stripTrailingZeros();
+
+        int scale = Math.max(currencyScale, amount.scale());
+        return new Money(amount, money.getCurrency(), scale);
     }
 
 }
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/domain/OrderItemPriceDetailImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/domain/OrderItemPriceDetailImpl.java
@@ -17,6 +17,7 @@
  */
 package org.broadleafcommerce.core.order.domain;
 
+import org.apache.commons.lang3.StringUtils;
 import org.broadleafcommerce.common.copy.CreateResponse;
 import org.broadleafcommerce.common.copy.MultiTenantCopyContext;
 import org.broadleafcommerce.common.currency.domain.BroadleafCurrency;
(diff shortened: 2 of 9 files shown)


common/src/main/java/org/broadleafcommerce/common/money/Money.java
```diff
@@ -220,6 +220,14 @@ public Money multiply(int amount) {
         return multiply(value);
     }
 
+    public Money multiplyWithRounding(int inInt, RoundingMode roundingMode) {
```
danielcolgrove, line 223: May make sense to have the multiply(int) call multiplyWithRounding(int, RoundingMode) passing HALF_EVEN as the default roundingMode.

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/domain/OrderItemPriceDetailImpl.java
```diff
@@ -180,20 +197,51 @@ public Money getFutureCreditAdjustmentValue() {
...
+        if (_roundingMode == null) {
+            ApplicationContext ctx = ApplicationContextHolder.getApplicationContext();
+            if (ctx != null) {
+                String modeStr = ctx.getEnvironment().getProperty(
+                        "item.offer.percent.rounding.mode");
+                if (StringUtils.isNotEmpty(modeStr)) {
+                    try {
+                        mode = RoundingMode.valueOf(modeStr);
+                    } catch (Exception e) {
+                        e.printStackTrace();
```
danielcolgrove: Should output user friendly message similar to what you did in PromotableItemFactoryImpl.java

common/src/main/java/org/broadleafcommerce/common/money/Money.java
danielcolgrove, line 223: Should change multiply(int) to call multiplyWithRounding(int, RoundingMode) passing the default rounding.  We'll only have the int multiply logic in one method.

danielcolgrove (changes requested):
In testing, used roundingMode=UP and scale=5.  Amounts shown in cart to 2 digits rounded up.  blc_order_item_dtl_adj shows adjusted_value with precision of 3.  Used Green Ghosts with $9.99, 50% off offer.  1 item in cart: $9.99 - $5.00 = $4.99.  blc_order_item_dtl_adj adjusted_value = $4.995.  This looks good.

Removed settings:
#item.offer.percent.rounding.mode=UP
#item.offer.percent.rounding.scale=5

Had a startup error:
Caused by: org.springframework.beans.TypeMismatchException: Failed to convert value of type 'java.lang.String' to required type 'java.lang.Integer'; nested exception is java.lang.NumberFormatException: For input string: "${item.offer.percent.rounding.scale}"
For input string: "${item.offer.percent.rounding.scale}"

common/src/main/java/org/broadleafcommerce/common/money/Money.java
danielcolgrove: Hmmm....this comment is a duplicate.  It disappeared on my review page so I re-entered it.

danielcolgrove:
The build failures are related to the startup error when the new config properties are missing.

common/src/main/java/org/broadleafcommerce/common/money/Money.java
bpolster: Purposely written to minimize potential impacts. The existing multiply logic has conditional behavior regarding scale that I did not evaluate.

danielcolgrove:
@bpolster Changes look good.  The build is failing with test errors:

Results :

Tests in error: 
  testApplyAllFulfillmentGroupOffersWithOrderItemOffers(org.broadleafcommerce.core.offer.service.processor.FulfillmentGroupOfferProcessorTest)
  testApplyItemQualifiersAndTargets(org.broadleafcommerce.core.offer.service.processor.ItemOfferProcessorTest)
  testFilterItemLevelOffer(org.broadleafcommerce.core.offer.service.processor.ItemOfferProcessorTest)
  testApplyAllItemOffers(org.broadleafcommerce.core.offer.service.processor.ItemOfferProcessorTest)
  testApplyAdjustments(org.broadleafcommerce.core.offer.service.processor.ItemOfferProcessorTest)
  testApplyOffersToOrder_Items(org.broadleafcommerce.core.offer.service.OfferServiceTest)
  testApplyOffersToOrder_Order(org.broadleafcommerce.core.offer.service.OfferServiceTest)

Tests run: 197, Failures: 0, Errors: 7, Skipped: 0

PR #2425, opened 2020-10-12
Cannot create a profile based on master profile

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/4209

### DIFF
--- a/admin/broadleaf-contentmanagement-module/src/main/java/org/broadleafcommerce/cms/field/domain/FieldGroupImpl.java
+++ b/admin/broadleaf-contentmanagement-module/src/main/java/org/broadleafcommerce/cms/field/domain/FieldGroupImpl.java
@@ -43,7 +43,8 @@ import javax.persistence.*;
  */
 @Entity
 @Inheritance(strategy = InheritanceType.JOINED)
-@Table(name = "BLC_FLD_GROUP")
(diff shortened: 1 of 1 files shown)


riteshadhikari17:
This change was fix for this issue https://github.com/BroadleafCommerce/QA/issues/4035. Need to find a solution that works for both cases.

StanislavFedorov:
@riteshadhikari17 
Added Unique Constraint for "NAME" and "SITE_DISC". Should work in both cases

riteshadhikari17:
There is a test failure. Happens locally as well. Here is the error
```
Unable to create unique key constraint (NAME, SITE_DISC) on table BLC_FLD_GROUP: database column 'SITE_DISC' not found. Make sure that you use the correct column name which depends on the naming strategy in use (it may not be the same as the property name in the entity, especially for relational types)
```

StanislavFedorov:
@riteshadhikari17 
Hmm...
"SITE_DISC" is the column from `SiteDiscriminatorImpl `class
For me this works locally. What is the difference with this enviroment?

riteshadhikari17:
It is running java 8. I Don't get error when I switch to java 11

StanislavFedorov:
I am on java 8
image

riteshadhikari17:
Are you running "mvn test"?

PR #2434, opened 2020-11-03
Update info on data driven enums on each metadata request, so in case of different sites data for enums should be for that site

oleksiimiroshnyk:
Fixes: BroadleafCommerce/QA#4225
Update info on data driven enums on each metadata request, so in case of different sites data for enums should be for that site


SunnyYu-98 (approved):
Tested locally and this fixes the issue

epenning (approved):
Code looks good!

PR #2437, opened 2020-11-05
add a test that checks OrderImpl.getHasOrderAdjustments()

gluckzhang:
**Add a test that checks OrderImpl.getHasOrderAdjustments()**

This test checks the return value of OrderImpl.getHasOrderAdjustments(). By default (when there is no order adjustment), the method should return false. If an adjustment is given, the method should return true.

**A Brief Overview**

I'd like to contribute in order to improve the test quality and effectiveness of Broadleaf. During mutation testing, I found that the behavior of [org.broadleafcommerce.core.order.domain.OrderImpl.getHasOrderAdjustments()](https://github.com/BroadleafCommerce/BroadleafCommerce/blob/develop-6.1.x/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/domain/OrderImpl.java#L781) is not specified. The method is covered by some of the tests. However, if the method body is replaced by `return true;`, there is no test fails.


riteshadhikari17:
@oleksiimiroshnyk Do you think this is ok to merge?

oleksiimiroshnyk:
@riteshadhikari17 think so..

PR #2440, opened 2020-11-11
Add support of uniqueConstraints to DirectCopyClassTransformer

oleksiimiroshnyk:
Add support of uniqueConstraints to DirectCopyClassTransformer
Remove SiteDisc annotation weave from FieldGroupImpl - it will be added in MT via xformTransformer, so it will add unique index/constraint for name and site_disc columns
Fixes: BroadleafCommerce/QA#4035


epenning (commented):
Code looks good!

The removal of the DirectCopyTransformMember from FieldGroupImpl is pretty straightforward.

The support for copying uniqueConstraints looks like it's related to https://github.com/BroadleafCommerce/MultiTenant-SingleSchema/pull/371, but that PR actually doesn't use it, since it makes a new index. So I'm guessing this is added since it's something that makes sense for us to support, is that right? Do you know if there's any existing use of this that could be used to test that it works as expected?

PR #2441, opened 2020-11-12
Fixed an issue where Product Add-on incorrectly shows "Included"

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/4245

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/call/ConfigurableOrderItemRequest.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/call/ConfigurableOrderItemRequest.java
@@ -44,7 +44,7 @@ public class ConfigurableOrderItemRequest extends AddToCartItem {
     protected boolean firstExpandable;
     protected boolean lastExpandable;
 
-    protected boolean priceIncludedInParent;
+    protected String pricingModelType;
 
     public Product getProduct() {
         return product;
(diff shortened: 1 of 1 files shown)


core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/call/ConfigurableOrderItemRequest.java
```diff
@@ -44,6 +44,8 @@
     protected boolean firstExpandable;
     protected boolean lastExpandable;
 
+    protected boolean priceIncludedInParent;
```
danielcolgrove: Shouldn't we just expose the PricingModelType?  That would be a more flexible approach should a customer need to add special logic on the frontend.

danielcolgrove (changes requested):

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/call/ConfigurableOrderItemRequest.java
StanislavFedorov: I could not find a way to get access to PricingModelType on Site. Because it is in ProductAddOnXref.

StanislavFedorov: Added `pricingModelType `to `ConfigurableOrderItemRequest` as a String

epenning (approved):
Looks good, matches up with Daniel's recommendation 👍

PR #2444, opened 2020-11-23
Added default locale for Site

troian88:
BroadleafCommerce/QA#3046
Now you can set the default locale per Site. Added locale selector on the general view of the Site in the admin panel.


riteshadhikari17:
Could you update this PR to target 6.2 instead of 6.1?

PR #2456, opened 2020-12-15
Replace const with var on js files to avoid minification issues

jetchegoyen:
BroadleafCommerce/QA#4294

Replace `const` with `var` on js files to avoid minification issues. 
As the current language to compile is ECMASCRIPT5, we cannot use some new features like const or let (since ES6).


troian88:
@jmitash Hello.
I tried to use this fix but received another error
Screenshot 2020-12-16 at 13 32 57
And one more thing. We have many 'const' declarations in the Enterprise module.


Maybe we should update the default ECMAScript version?

troian88 (approved):

PR #2465, opened 2021-01-04
Deproxy nested fields when reading field values

cade-rea:
When viewing a collection item in the admin, nested object fields are not read properly when loading the Entity for display. For instance, when viewing a CustomerAddress, the phone number fields do not populate with the saved data.

During the resolution of a field path (eg `address.phonePrimary.phoneNumber`), this patch will check if an object is a HibernateProxy and deproxy it. This allows the values from that object to populate the entity form.

QA Issue [4313](https://github.com/BroadleafCommerce/QA/issues/4313)


cja769 (approved):

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/module/FieldManager.java
```diff
@@ -94,6 +95,10 @@ public Object getFieldValue(Object bean, String fieldName) throws IllegalAccessE
             if (field != null) {
                 field.setAccessible(true);
                 value = field.get(value);
+                
+                if (value instanceof HibernateProxy) {
```
danielcolgrove: No need to add this if check.  The HibernateUtils#deproxy() already does the check.

danielcolgrove (changes requested):

danielcolgrove (approved):

PR #2467, opened 2021-01-12
Decode validation string when xss protection is enabled

SunnyYu-98:
Decode product option validation string during product option value regex evaluation

**A Brief Overview**
Fixes https://github.com/BroadleafCommerce/QA/issues/4324

**Additional context**
HelpScout: https://secure.helpscout.net/conversation/1392821250/44564?folderId=4164930

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/ProductOptionValidationServiceImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/ProductOptionValidationServiceImpl.java
@@ -32,6 +32,7 @@ import org.broadleafcommerce.core.order.service.type.MessageType;
 import org.broadleafcommerce.core.workflow.ActivityMessages;
 import org.owasp.esapi.ESAPI;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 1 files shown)


core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/ProductOptionValidationServiceImpl.java
```diff
@@ -67,7 +67,9 @@ public Boolean validate(ProductOption productOption, String value) {
             throw new RequiredAttributeNotProvidedException(message, attributeName);
         } else {
             String validationString = productOption.getValidationString();
+            validationString = Boolean.parseBoolean(environment.getProperty("exploitProtection.xssEnabled", "false")) ? ESAPI.encoder().decodeForHTML(validationString) : validationString;
```
danielcolgrove: Let's pull the environment out to a property with an @Value annotation.  I believe environment.getProperty will force a propagated call to resolve the property each time.  So a DB call, property check, etc.  The exploitProtection.xssEnabled should not be changing at runtime so a once and done at class initialization should be good.

danielcolgrove (changes requested):

danielcolgrove (approved):

PR #2469, opened 2021-01-14
Update XSS Filter and Wrapper strip XSS logic

SunnyYu-98:
**Main Changes**
- Update `stripXSS` logic to use `ESAPI.encoder().encodeForHTML` instead of `ESAPI.validator(). getValidSafeHTML` so that input isn't completely removed
- Update `XssFilter` and `XssRequestWrapper` for easier client override

**A Brief Overview**
Fixes https://github.com/BroadleafCommerce/QA/issues/4328

**Additional context**
As I was providing the override classes for FlyWire as a hot fix, it was not easy to override the methods that I needed because they were all private. That's why I made the changes to make those methods protected.

`ESAPI.encoder().encodeForHTML` vs. `ESAPI.validator(). getValidSafeHTML`: 
- `encodeForHTML` simply encodes every special characters in user input, but `getValidSafeHTML` completely removes all input that is a XSS threat. Because of that, FlyWire was not able to use `\` in their product option input

### DIFF
--- a/common/src/main/resources/esapi/validation.properties
+++ b/common/src/main/resources/esapi/validation.properties
@@ -2,7 +2,7 @@
 # #%L
 # BroadleafCommerce Common Libraries
 # %%
-# Copyright (C) 2009 - 2016 Broadleaf Commerce
+# Copyright (C) 2009 - 2020 Broadleaf Commerce
 # %%
 # Licensed under the Broadleaf Fair Use License Agreement, Version 1.0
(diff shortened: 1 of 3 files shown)


SunnyYu-98:
Ran the following new approach by Jerry, PR updated accordingly:
- Use `ESAPI.validator().getValidInput` first using the custom validator rule and have the original logic as fall back

Issue #2472, opened 2021-01-18
Sku inconsistent caused by concurrent checkout


project-concerto:
# Problem
The inventory can be inconsistent when multiple customers purchase the same product concurrently. This is because that read and modification to sku during checkout is not atomic.
# How It Happened
In [DefaultPaymentGatewayCheckoutService#initiateCheckout](https://github.com/BroadleafCommerce/BroadleafCommerce/blob/941a20c29d55283991c84fa5a88674f40937a4ef/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/payment/service/DefaultPaymentGatewayCheckoutService.java#L252), there is no transaction defined to protect the whole process of checkout. Therefore, the SQL statements to read and update BLC_SKU items are not atomic and can be interleaved with others. The SQL statements of interest is shown belosw.

```sql
# begin checkout
...
# this is an auto commit query which read inventory(BLC_SKU.QUANTITY) of product
# result is used to validate whether inventory is available in InventoryServiceImpl#decrementSKU#L209
select * from BLC_PRODUCT productimp0_  
    left join ...
    left outer join BLC_SKU skuimpl3_ on productimp0_.DEFAULT_SKU_ID=skuimpl3_.SKU_ID
    left join ...
where productimp0_.PRODUCT_ID=6  
...
begin # this statement is used to update inventory
update BLC_SKU set ...QUANTITY_AVAILABLE=9998... where SKU_ID=6
commit
begin # complete the order
update BLC_ORDER set ... ORDER_STATUS='SUBMITTED'... where ORDER_ID=88
commit
# finish checkout
```
Assuming that customer 1 and customer 2 buy product A at the same time, there may exist the following interleaving which would make inventory inconsistent (There is only one product A, but both customer 1 and customer 2 bought it successfully).
| Customer 1 buys product A   | Customer 2 buys product A  |
|  ----  | ----  |
| Read A.SKU.QUANTITY => 1  |  |
|   | Read A.SKU.QUANTITY => 1 |
| Update A.SKU.QUANTITY <= 0  |  |
|   | Update A.SKU.QUANTITY <= 0 |

# Possible Solution
[...]

PR #2473, opened 2021-01-18
use serializable transaction wrap the process of checkout

project-concerto:
See #2472


danielcolgrove:
See https://github.com/BroadleafCommerce/BroadleafCommerce/issues/2472

PR #2486, opened 2021-02-01
Fixed an issue where Customer updated on registration validation errors

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/4356


core/broadleaf-framework-web/src/main/java/org/broadleafcommerce/core/web/controller/account/BroadleafRegisterController.java
```diff
@@ -114,6 +114,8 @@ public String processRegister(RegisterCustomerForm registerCustomerForm, Binding
             }
             return StringUtils.isBlank(redirectUrl) ? getRegisterSuccessView() : "redirect:" + redirectUrl;
         } else {
+            //We need to detach Customer to avoid update Customer in DB QA#4356
```
oleksiimiroshnyk: guess, worth mentioning in the comment that customer was read from the db and is "attached to session", later CustomerStateFilter does order "unlock", where it starts transaction, and so on transaction commit session is flashed and changes to customer is persisted. for this case we need to detach customer from session. or somethign like that

oleksiimiroshnyk (approved):

PR #2504, opened 2021-04-06
Update default decimal value for locale

oleksiimiroshnyk:
- Update default decimal value for locale
BroadleafCommerce/QA#4410


danielcolgrove:
@oleksiimiroshnyk This approach will not work.  My recreation instructions were incomplete.  Multiple users across the various countries will all be working at the same time.  Icon uses multiple tabs - one for each country.  With this approach the locale set on the last call to buildPropertiesFromPolymorphicEntities is the only locale that will work.  The approach needs to be able to support multiple locales running at once.

oleksiimiroshnyk:
@danielcolgrove not quite following why this approach will not work.
This piece of code will be executed on each request, so if locale is correctly determined from the request that code will update default value in metadata for this request based on locale determined from the request. 
I tested:
opened a test site: locale is En, pressed add Order, then go back to list, switched locale to Fr and pressed add Order, switched back to En.
opened another tab with global admin, swithced locale to Fr, switched to test site, tried to add Order
in prev tab(with test siste and locale En) pressed add Order
I tried this couple of times in different combinations, looks like working fine.
Do you mean another some steps with multiple tabs ?

danielcolgrove:
It failed when I tested with the Icon code.  For Icon site is loaded in a separate tab.  I'll try to recreate with our code.  I do see on line 766 that the cache is referenced.  Perhaps that cache is flushed with we switch locale but that is not happening when they switch tabs.

danielcolgrove:
This does seem to be working now.

PR #2508, opened 2021-04-14
add versions to css, js files that are referenced directly

oleksiimiroshnyk:
- add versions to css, js files that are referenced directly
Fixes: BroadleafCommerce/QA#4418


danielcolgrove (approved):

danielcolgrove:
I approved but we should put a version on all the static files.  Even the BLC files.  If they are using bundles with versioning enabled, it's pointless.  But, if they aren't then the versions would be needed.

oleksiimiroshnyk:
@danielcolgrove pushed a change to bundle processor, so in case of bundle is off files that are in bundle will get some random param:
 xxx.js?v=0.5234524 aaa.css?v=0.23241234

PR #2515, opened 2021-05-07
Add validation for deletion of categories, check if it is a primary category for some product

oleksiimiroshnyk:
- Add validation for deletion of categories, check if it is a primary category for some product
Fixes: BroadleafCommerce/QA#4441


admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/server/service/handler/CategoryCustomPersistenceHandler.java
```diff
@@ -176,6 +188,21 @@ public Entity update(PersistencePackage persistencePackage, DynamicEntityDao dyn
         }
     }
 
+    @Override
+    public void remove(PersistencePackage persistencePackage, DynamicEntityDao dynamicEntityDao, RecordHelper helper) throws ServiceException {
+        Property id = persistencePackage.getEntity().getPMap().get("id");
+        Query query = dynamicEntityDao.getStandardEntityManager().createQuery("SELECT xref FROM org.broadleafcommerce.core.catalog.domain.CategoryProductXrefImpl xref" +
```
danielcolgrove: Can we pull this out into the Category DAO?  I know we have a few queries in a couple CPH's but I think that is bad practice.

danielcolgrove (changes requested):

danielcolgrove (approved):

PR #2532, opened 2021-06-18
Entity lookups are displaying empty fields when participating in an adornedTargetCollection relationship

danielcolgrove:
This is a fix to not show the empty fields if they don't naturally appear in the entity property list.  Fixes https://github.com/BroadleafCommerce/QA/issues/4473

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/service/FormBuilderServiceImpl.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/service/FormBuilderServiceImpl.java
@@ -504,7 +504,7 @@ public class FormBuilderServiceImpl implements FormBuilderService {
                 selectize = true;
 
                 Property p = cmd.getPMap().get(atcmd.getSelectizeVisibleField());
-                if (p != null) {
+                if (propertyExistsInResultSet(p, drs)) {
                     BasicFieldMetadata md = (BasicFieldMetadata) p.getMetadata();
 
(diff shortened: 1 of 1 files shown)


admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/service/FormBuilderServiceImpl.java
```diff
@@ -514,7 +514,7 @@ public ListGrid buildCollectionListGrid(String containingEntityId, DynamicResult
             } else {
                 for (String fieldName : atcmd.getGridVisibleFields()) {
                     Property p = cmd.getPMap().get(fieldName);
-                    if (p != null) {
+                    if (propertyExistsInResultSet(p, drs)) {
```
oleksiimiroshnyk, line 517: if property doesn't exist on entity I would think that it is not in class metadata and so cmd.getPMap.get(fieldName) should return null?
If we query less fields than in metadata this fix should address it, but I don't think we do this OOB ?

oleksiimiroshnyk (approved):

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/service/FormBuilderServiceImpl.java
danielcolgrove: I looked for the possibility of using the class metadata.  Unfortunately, the class metadata "inherits" some properties from the adornedTargetCollection and in this case it has the adorned fields.

PR #2544, opened 2021-07-07
Update solr to 8.9.0

oleksiimiroshnyk:
Fixes: BroadleafCommerce/QA#4491


danielcolgrove (approved):
Looks good.  I wasn't sure if a simple update of the version would be enough or if there were SolrJ updates needed for searching or indexing.

PR #2553, opened 2021-08-12
Ensure order lock can be released

project-concerto:
**Performcheckout order lock**
- Fixes the issue that order lock may be not released finally

**Brief Overview**
move validation code into try-finally block to ensure that lock will always be released correctly

**Labels:**
 - Bug or Enhancement


riteshadhikari17:
Branch 6.1 Is no longer in active development.

PR #2556, opened 2021-08-12
Added Catalog selector for the Product duplication

troian88:
Added catalog selector

**A Brief Overview**
Added ability for select catalog destination while product duplication

**Link to QA issue**
https://github.com/BroadleafCommerce/QA/issues/4483

**Add Labels to the right panel and delete this list:**
 - Enhancement
 
**Add Milestone to the right panel and delete this section (required):**
6.1.8-GA

**Additional context**
https://docs.google.com/document/d/1XolbYtJ00aYsEwZPqSYJ0j-KK2DE9JjacWuPq4mvOoI/edit

### DIFF
--- a/admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/web/controller/entity/AdminOfferController.java
+++ b/admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/web/controller/entity/AdminOfferController.java
@@ -117,12 +117,13 @@ public class AdminOfferController extends AdminBasicEntityController {
      */
     @Deprecated
     @Override
-    @RequestMapping(value = "/{id}/duplicate", method = RequestMethod.POST)
+    @RequestMapping(value = "/{id}/duplicate/{catalogId}", method = RequestMethod.POST)
     public String duplicateEntity(HttpServletRequest request, HttpServletResponse response, Model model,
-            @PathVariable  Map<String, String> pathVars, @PathVariable(value="id") String id,
-            @ModelAttribute(value="entityForm") EntityForm entityForm, BindingResult result) 
+                                  @PathVariable Map<String, String> pathVars, @PathVariable(value = "id") String id,
+                                  @PathVariable(value = "catalogId") final String catalogId,
+                                  @ModelAttribute(value = "entityForm") EntityForm entityForm, BindingResult result)
             throws Exception {
-        return super.duplicateEntity(request, response, model, pathVars, id, entityForm, result);
+        return super.duplicateEntity(request, response, model, pathVars, id, catalogId, entityForm, result);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/controller/entity/AdminBasicEntityController.java
```diff
@@ -417,24 +421,25 @@ public String addEntity(HttpServletRequest request, HttpServletResponse response
         return "ajaxredirect:" + getContextPath(request) + sectionKey + "/" + entity.getPMap().get("id").getValue();
     }
 
-    @RequestMapping(value = "/{id}/duplicate", method = RequestMethod.POST)
+    @RequestMapping(value = "/{id}/duplicate/{catalogId}", method = RequestMethod.POST)
```
oleksiimiroshnyk, line 424: I think we need leave as it is and all "catalog" related changes should got to MT module.
for example there is MultiTenantAdminBasicEntityControllerExtensionHandler where we can adjust url of action to trigger something in MultiTenantAdminBasicEntityController(it overrides AdminBasicEntityController )

common/src/main/java/org/broadleafcommerce/common/persistence/EntityDuplicatorImpl.java
```diff
@@ -136,7 +138,7 @@ public boolean validate(final Object entity) {
 
     @Transactional(TransactionUtils.DEFAULT_TRANSACTION_MANAGER)
     @Override
-    public <T> T copy(final Class<T> entityClass, final Long id) {
+    public <T> T copy(final Class<T> entityClass, final Long id, final String catalogIdentifier) {
```
oleksiimiroshnyk, line 141: think we can leave as it is and do  BroadleafRequestContext.getBroadleafRequestContext().getAdditionalProperties().put("duplicateToCatalogId", catalogIdentifier); in MultiTenantAdminBasicEntityController for example.
Logic is that some clients don't use MT module but will have logic belonging to MT and this will not work I think

common/src/main/java/org/broadleafcommerce/common/persistence/EntityDuplicatorImpl.java
```diff
@@ -193,7 +196,7 @@ public boolean validate(final Object entity) {
 
     @Transactional(TransactionUtils.DEFAULT_TRANSACTION_MANAGER)
     @Override
-    public <T> T copy(final MultiTenantCopyContext context, final MultiTenantCloneable<T> entity) {
+    public <T> T copy(final MultiTenantCopyContext context, final MultiTenantCloneable<T> entity, final Long catalogIdentifier) {
```
oleksiimiroshnyk, line 199: looks like this is not used at all

admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/web/controller/entity/AdminOfferController.java
```diff
@@ -117,12 +117,13 @@ public String addEntity(HttpServletRequest request, HttpServletResponse response
      */
     @Deprecated
     @Override
-    @RequestMapping(value = "/{id}/duplicate", method = RequestMethod.POST)
+    @RequestMapping(value = "/{id}/duplicate/{catalogId}", method = RequestMethod.POST)
```
oleksiimiroshnyk, line 120: I think we can have a new controller in MT module that will have method with such signature

oleksiimiroshnyk (changes requested):

PR #2572, opened 2021-09-07
Added the ability for DirectCopyClassTransformer to replace a class-level @Cache annotation

danielcolgrove:
Fixes https://github.com/BroadleafCommerce/QA/issues/4522

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/extensibility/jpa/copy/DirectCopyClassTransformer.java
+++ b/common/src/main/java/org/broadleafcommerce/common/extensibility/jpa/copy/DirectCopyClassTransformer.java
@@ -202,7 +202,6 @@ public class DirectCopyClassTransformer extends AbstractClassTransformer impleme
                     ClassFile templateFile = template.getClassFile();
                     ConstPool constantPool = classFile.getConstPool();
                     buildClassLevelAnnotations(classFile, templateFile, constantPool);
-                    buildClassLevelCacheAnnotation(classFile, templateFile, constantPool);
 
                     // Copy over all declared fields from the template class
                     // Note that we do not copy over fields with the @NonCopiedField annotation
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/broadleafcommerce/common/extensibility/jpa/copy/DirectCopyClassTransformer.java
```diff
@@ -558,6 +560,83 @@ protected Annotation getEntityListeners(ConstPool constantPool, Annotation exist
...
+                Object object = itr.next();
+                if (AnnotationsAttribute.class.isAssignableFrom(object.getClass())) {
+                    AnnotationsAttribute attr = (AnnotationsAttribute) object;
+                    Annotation[] items = attr.getAnnotations();
+                    for (Annotation annotation : items) {
+                        String typeName = annotation.getTypeName();
+                        if (typeName.equals(javax.persistence.Cache.class.getName())
+                                || typeName.equals(org.hibernate.annotations.Cache.class.getName())) {
+                            logger.debug("Stripping out previous Cache annotation at the class level - will merge into new EntityListeners");
+                            //annotationsAttribute.addAnnotation(templateCache);
```
oleksiimiroshnyk: guess we can clear commented lines?

oleksiimiroshnyk (approved):
Few questions:
why do we have a new method and change of existing that processes class level annotation? Cache is also class level annotation..
We don't stop loop once we find cache annotation, do we expect >1 cache annotation and we are interested only in the last one ?
And what about if we will have to Cache annotations one from javax package and another from  org.hibernate ? Not sure if this is a valid case, but just thoughts

common/src/main/java/org/broadleafcommerce/common/extensibility/jpa/copy/DirectCopyClassTransformer.java
danielcolgrove: Fixed this

danielcolgrove:
Q) why do we have a new method and change of existing that processes class level annotation? Cache is also class level annotation.
A) The method name (buildClassLevelAnnotations) is poorly names.  It actually only updates EntityListeners.  I've changed the methods creating a listener method and a cache method which gets called from buildClassLevelAnnotations.  In 6.1 there is also an Index method.
Q) We don't stop loop once we find cache annotation, do we expect >1 cache annotation and we are interested only in the last one ?
A) Good catch, the logic now breaks
Q) And what about if we will have to Cache annotations one from javax package and another from org.hibernate ? Not sure if this is a valid case, but just thoughts
A) I thought about that but all we have currently is Hibernate.  We can leave it as is for now.

PR #2575, opened 2021-09-14
Check property on null while building Collection List Grid

troian88:
**Add proper title**
Fixes BroadleafCommerce/QA#4527 NPE issue while building Collection list grid

**Add Labels to the right panel and delete this list:**
 - Bug

**Additional context**
https://secure.helpscout.net/conversation/1629914689/47679?folderId=305491


admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/service/FormBuilderServiceImpl.java
```diff
@@ -731,6 +731,7 @@ public ListGrid buildCollectionListGrid(String containingEntityId, DynamicResult
     }
 
     protected boolean propertyExistsInResultSet(Property property, DynamicResultSet drs) {
+        if(property == null) return false;
```
oleksiimiroshnyk: I think we prefer another code style
if (property == null) {
  return false;
}

PR #2579, opened 2021-10-04
Don't fetch primary sku media during product list grid load

oleksiimiroshnyk:
- Don't fetch primary sku media during product list grid load
Fixes: BroadleafCommerce/QA#4535


riteshadhikari17:
@Jerry77oz is this ok to merge?

PR #2599, opened 2021-11-24
Fix duplicator name generator, now it correctly iterates over copy numbers

oleksiimiroshnyk:
- Fix duplicator name generator, now it correctly iterates over copy numbers

Fixes: BroadleafCommerce/QA#4580


riteshadhikari17:
I reverted the change for URL. I will merge this fix for name and we can open new PR for it. BTW `original.getUrl()` doesn't return null because in `ProductImpl` there is logic to generate URL if it's null.

PR #2606, opened 2021-12-06
Moved filtering by locate to sql Query in readNamedOrderFor

KatrukOV:
Logic of filtering by locate was implemented via java code.
https://github.com/BroadleafCommerce/QA/issues/4591

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/dao/OrderDaoImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/dao/OrderDaoImpl.java
@@ -315,16 +315,14 @@ public class OrderDaoImpl implements OrderDao {
     @Override
     @SuppressWarnings("unchecked")
     public Order readNamedOrderForCustomer(final Customer customer, final String name) {
-        final Query query = em.createNamedQuery("BC_READ_NAMED_ORDER_FOR_CUSTOMER");
+        final Query query = em.createNamedQuery("BC_READ_NAMED_ORDER_FOR_CUSTOMER_WITH_LOCALE");
+
+        final Locale locale = BroadleafRequestContext.getBroadleafRequestContext().getLocale();
+        query.setParameter("locale", locale);
         query.setParameter("customerId", customer.getId());
         query.setParameter("orderStatus", OrderStatus.NAMED.getType());
         query.setParameter("orderName", name);
-        if (BroadleafRequestContext.getBroadleafRequestContext() != null) {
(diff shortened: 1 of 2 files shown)


core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/dao/OrderDaoImpl.java
```diff
@@ -319,27 +319,21 @@ public Order readNamedOrderForCustomer(final Customer customer, final String nam
...
-        // Filter out orders that don't match the current locale (if one is set)
         if (BroadleafRequestContext.getBroadleafRequestContext() != null) {
-            ListIterator<Order> iter = orders.listIterator();
-            while (iter.hasNext()) {
-                Locale locale = BroadleafRequestContext.getBroadleafRequestContext().getLocale();
-                Order order = iter.next();
-                if (locale != null && !locale.equals(order.getLocale())) {
-                    iter.remove();
-                }
+            final Locale locale = BroadleafRequestContext.getBroadleafRequestContext().getLocale();
```
oleksiimiroshnyk: I think this needs a little bit more, see the named query BC_READ_NAMED_ORDER_FOR_CUSTOMER that is in config/bc/jpa/domain/Order.orm.xml:64
Guess we need to add something like this to that query
AND (:locale is null or order.locale = :locale)

or maybe just create a new named query same as above and accepts locale as parameter and use it in case  BroadleafRequestContext.getBroadleafRequestContext().getLocale() != null ?

oleksiimiroshnyk (approved):

StanislavFedorov (approved):

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/dao/OrderDaoImpl.java
danielcolgrove: The Java logic should still be honored.  If there is not a BRC locale, don't filter on locale.  If there is a BRC locale, use it.

KatrukOV: AND (:locale is null or order.locale = :locale)
This construct provides filtering if locale is present (not null) and ignored if locate is null.

KatrukOV: I have checked both options (locale = null, and not null) and both work fine.

danielcolgrove (approved):

PR #2613, opened 2021-12-13
Upgraded log4j and excluded from xmlbeans for 5.2

riteshadhikari17:
**A Brief Overview**
Upgrade the log4j. Needed to exclude in XMLBeans to make sure latest version will get used.

[QA Link](https://github.com/BroadleafCommerce/QA/issues/4606)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -768,7 +768,7 @@
             <dependency>
                 <groupId>org.apache.logging.log4j</groupId>
                 <artifactId>log4j-bom</artifactId>
-                <version>2.11.1</version>
+                <version>2.15.0</version>
                 <scope>import</scope>
                 <type>pom</type>
                 <optional>true</optional>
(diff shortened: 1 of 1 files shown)


oleksiimiroshnyk (approved):

danielcolgrove (approved):

oleksiimiroshnyk:
@riteshadhikari17 I see we still have:
```
            <dependency>
                <groupId>log4j</groupId>
                <artifactId>log4j</artifactId>
                <version>1.2.17</version>
                <optional>true</optional>
            </dependency>
```
do we need to remove it ? or maybe upgrade ?

riteshadhikari17:
We have been supporting both log4j and log4j2. I had mentioned it before we released 6.2 that we should drop log4j but we got busy and didn't do it. @danielcolgrove Since no one is on Production with 6.2, is it ok to remove log4j in 6.2? Everyone who is trying 6.2 or looking to migrate to 6.2 are probably not using it. That 1.2.17 is from 2012 and has vulnerabilities. We can say because of log4shell we dropped log4j and upgraded log4j2 to latest release. This will be good point to include in the Next 6.2 release. We can leave 6.1 and below as is.

danielcolgrove:
Most everyone using 6.2 is using SpringBoot which is using LogBack.  We can remove log4j from 6.2.

riteshadhikari17:
Cool. Here is ticket for that. https://github.com/BroadleafCommerce/QA/issues/4626

PR #2615, opened 2021-12-15
Removed redundant concurrency protection

KatrukOV:
Since we already have an order blocking in CartStateFilter and we are using DatabaseOrderLockManager, there is no need to use this block in CheckoutServiceImpl.performCheckout()

Link to QA
[BroadleafCommerce/QA#4590](https://github.com/BroadleafCommerce/QA/issues/4590)


StanislavFedorov (approved):

oleksiimiroshnyk (commented):
think we need also remove 
 finally {
                this.removeLock(order.getId());
            }

PR #2618, opened 2021-12-22
Added synchronized for createNewCartForCustomer

KatrukOV:
When creating an order in OrderServiceImpl.createNewCartForCustomer(), a sync block was added to ensure that one order is created for one customer.

Link to QA
[BroadleafCommerce/QA#4589](https://github.com/BroadleafCommerce/QA/issues/4589)


StanislavFedorov (approved):

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/OrderServiceImpl.java
```diff
@@ -214,7 +212,10 @@
     @Override
     @Transactional("blTransactionManager")
     public Order createNewCartForCustomer(Customer customer) {
-        return orderDao.createNewCartForCustomer(customer);
+        final Object look = Objects.isNull(customer.getId()) ? customer : customer.getId();
```
danielcolgrove: Should this variable be called "lock" instead of "look"?

PR #2624, opened 2022-01-12
Removed redundant concurrency protection in CheckoutServiceImpl

KatrukOV:
Removed lock mechanism in CheckoutServiceImpl.

Link to QA issue
[QA-4590](https://github.com/BroadleafCommerce/QA/issues/4590)

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/checkout/service/CheckoutServiceImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/checkout/service/CheckoutServiceImpl.java
@@ -51,12 +51,6 @@ public class CheckoutServiceImpl implements CheckoutService {
 
     @Resource(name="blOrderService")
     protected OrderService orderService;
-    
-    /**
-     * Map of locks for given order ids. This lock map ensures that only a single request can handle a particular order
-     * at a time
-     */
-    protected static ConcurrentMap<Long, Object> lockMap = new ConcurrentHashMap<>();
 
     @Override
     public CheckoutResponse performCheckout(Order order) throws CheckoutException {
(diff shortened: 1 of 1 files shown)


oleksiimiroshnyk (approved):

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/checkout/service/CheckoutServiceImpl.java
```diff
@@ -104,24 +95,4 @@ protected boolean hasOrderBeenCompleted(Order order) {
...
     }
 
-    /**
-    * Get an object to lock on for the given order id
-    * 
-    * @param orderId
-    * @return null if there was not already a lock object available. If an object was already in the map, this will return
-    * that object, which means that there is already a thread attempting to go through the checkout workflow
-    */
-    protected Object putLock(Long orderId) {
```
danielcolgrove, line 114: @oleksiimiroshnyk @KatrukOV Should we leave these methods and deprecate them for now?  I would be cautious knowing that there is a remote possibility that a customer has the performCheckout method overridden with code that still references these methods.

oleksiimiroshnyk: We thought about it and decided to go this way. We can deprecate in 6.1 and remove in 6.2. But I think better to let code fail now rather than rely on something that is not really needed/expected.
Let us know if we should remove in 6.1 or 6.2

danielcolgrove: Let's deprecate in 6.1 and remove in 6.2.

PR #2630, opened 2022-01-17
Remove log4j 1

oleksiimiroshnyk:
- Remove log4j 1

Fixes: BroadleafCommerce/QA#4626


common/src/main/java/org/broadleafcommerce/common/cache/Log4jStatisticsServiceLogAdapter.java
```diff
@@ -17,8 +17,10 @@
...
 package org.broadleafcommerce.common.cache;
 
-import org.apache.log4j.Level;
-import org.apache.log4j.LogManager;
+
+import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.core.config.Configurator;
 
 /**
```
riteshadhikari17: There is already a class Log4j2StatisticsServiceLogAdapter. Can we just delete this one?

riteshadhikari17 (commented):
There are log4j.xml files in tests. Can we delete those if we don't use them? If we do use them, we should probably update tests to use log4j2.xml. These are the files
https://github.com/BroadleafCommerce/BroadleafCommerce/blob/develop-6.2.x/common/src/test/resources/log4j.xml
https://github.com/BroadleafCommerce/BroadleafCommerce/blob/develop-6.2.x/integration/src/test/resources/log4j.xml

oleksiimiroshnyk:
@riteshadhikari17 I reworked log4j.xml to log4j2.xml, looks like works ok

riteshadhikari17 (approved):

PR #2647, opened 2022-03-21
Remove inline scripts in scrollbar script so CSP should not react on it

MarekB01:
I have deleted this function "oncontextmenu='return false;'" everywhere. Also i have added new functional.

Fixed an issue of getting rid of inline scripts

Now Content Security Policy do not sending a report related to loading a resource at inline ("script-src")

BroadleafCommerce/QA#4655

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/lib/plugins/jquery.mCustomScrollbar.js
+++ b/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/lib/plugins/jquery.mCustomScrollbar.js
@@ -73,13 +73,6 @@ plugin home: http://manos.malihu.gr/jquery-custom-content-scroller
 				if($.support.touch){
 					mCSB_container.addClass("mCS_touch");
 				}
-				$(mCSB_container).contextmenu(function() {
-					return false;
-				});
-				$(mCSB_scrollTools).contextmenu(function() {
-					return false;
-				});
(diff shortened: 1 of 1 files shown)


admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/lib/plugins/jquery.mCustomScrollbar.js
```diff
@@ -73,7 +73,13 @@ plugin home: http://manos.malihu.gr/jquery-custom-content-scroller
 				if($.support.touch){
 					mCSB_container.addClass("mCS_touch");
 				}
-				mCSB_container.after("<div class='mCSB_scrollTools' style='position:absolute;'><div class='mCSB_draggerContainer'><div class='mCSB_dragger' style='position:absolute;' oncontextmenu='return false;'><div class='mCSB_dragger_bar' style='position:relative;'></div></div><div class='mCSB_draggerRail'></div></div></div>");
+				$(mCSB_container).contextmenu(function() {
```
oleksiimiroshnyk: guess this is not equivalent of what was before...
context menu event is attached to div with class mCSB_dragger not $mCSB_container

oleksiimiroshnyk:
decided to simply wipe this away because context menu is shown on right click everywhere, so this was not really working at all. And it is easier to get ride of it

oleksiimiroshnyk (approved):

PR #2648, opened 2022-03-21
Revisit Recursive Relationships for UpSaleProduct

KatrukOV:
Created UpSaleProductCustomPersistenceHandler with validateUpSaleProduct.

Link to QA issue
[QA-4643](https://github.com/BroadleafCommerce/QA/issues/4643)

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/entityForm.js
+++ b/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/entityForm.js
@@ -712,6 +712,11 @@ $(document).ready(function() {
                         }
                     });
 
+                    let validationErrors = $(data).find('.fieldError');
+                    if (validationErrors.length !== 0) {
+                        BLCAdmin.showMessageAsModal(BLCAdmin.messages.validationError, validationErrors.text());
+                    }
+
                     var $actions = BLCAdmin.currentModal().find('.entity-form-actions');
                     $actions.find('button').show();
                     $actions.find('img.ajax-loader').hide();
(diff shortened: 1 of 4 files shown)


StanislavFedorov (approved):

danielcolgrove:
We need to make sure that we surface good error messages to that uses know that is wrong.  I tested:

Product A -> upsale to Product B
Product B -< upsale to Product A

The error message I get was simply "A validation error occurred on the server".  There is nothing actionable with that error message.
Screen Shot 2022-03-21 at 8 29 19 AM

admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/entityForm.js
```diff
@@ -712,6 +712,11 @@ $(document).ready(function() {
                         }
                     });
 
+                    let validationErrors = $(data).find('.fieldError');
```
oleksiimiroshnyk: use var instead of let, we have old minification engine that complains about using let

oleksiimiroshnyk (approved):

danielcolgrove:
Not to make this tougher but could we get the name of the product that causes the recursion?

At a minimum, I would change the error to:

This linked upSale product cannot be used because it causes a recursive relationship with another product

danielcolgrove:
It works best with a simple recursion: Product A -> Product B -> Product A.  When you have three products in recursion calling out the recursive product isn't as clear.  Product A -> Product B -> Product C -> Product A - when trying to make that last association between Product C and Product A I get a validation error that I have a Recursion with Product B.  That is true but could be hard for a user to understand where exactly that is coming from.  I think we leave it like this.  Thanks for taking the extra time to get this enhanced product.

oleksiimiroshnyk (approved):

PR #2656, opened 2022-04-04
Fixed ampersand issue for offer rules

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/4663

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/offer/service/processor/AbstractBaseProcessor.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/offer/service/processor/AbstractBaseProcessor.java
@@ -47,19 +47,10 @@ import org.broadleafcommerce.core.order.service.type.FulfillmentType;
 import org.broadleafcommerce.profile.core.domain.Customer;
 import org.joda.time.DateTimeZone;
 import org.joda.time.LocalDateTime;
-import java.util.ArrayList;
-import java.util.Calendar;
-import java.util.Collection;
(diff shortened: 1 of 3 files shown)


riteshadhikari17:
There was something wrong with build so I have kicked off another build manually. Also could you revert the wild card import?

StanislavFedorov:
reverted import

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/offer/service/processor/AbstractBaseProcessor.java
```diff
@@ -279,6 +283,18 @@ protected boolean couldOrderItemMeetOfferRequirement(OfferItemCriteria criteria,
 
         return appliesToItem;
     }
+
```
danielcolgrove, line 282: Would it make sense to pull this out into a helper class?  Are there other places we could use this logic.  If offer rules have this issue, what about Customer Segment, Targeters, other rule builders?

StanislavFedorov:
>Would it make sense to pull this out into a helper class?

This sounds reasonable. I overlooked this.
The current solution is to move this "escapeSpecialCharacters" logic to the general class for all rule builders.
The changes looked pretty simple but I had to test all cases and implement 'unescapeSpecialCharacters'. So it took more time..
Because now we store that Ampersand in DB as &amp; , but before this, we just did the conversion for Offer rules "on the fly"
I checked this for Offers rules and for Customer Segment/Targeters

riteshadhikari17 (approved):

PR #2660, opened 2022-04-08
Fix adorned list grids selections

oleksiimiroshnyk:
- Fix adorned list grids selections

Fixes: BroadleafCommerce/QA#4671


riteshadhikari17:
This works but there is another use case that is failing. If there is validation error, it fails again. Here are the steps:

Follow steps in QA ticket from 1-3. First pick ABC corp and click add. This should bring validation error. Close error then pick XYZ corp. This also shows the same validation error

riteshadhikari17 (approved):

PR #2664, opened 2022-04-14
Change the condition to evaluate if classExcludeRegexList is used

riteshadhikari17:
**A Brief Overview**
classExcludeRegexList is not thread safe so changes were made which was causing unnecessary noisy logging. So changed the condition to evaluate if old list was used.

For https://github.com/BroadleafCommerce/QA/issues/4681

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/copy/MultiTenantCopier.java
+++ b/common/src/main/java/org/broadleafcommerce/common/copy/MultiTenantCopier.java
@@ -202,10 +202,8 @@ public abstract class MultiTenantCopier implements Ordered {
     @Deprecated
     protected Boolean excludeFromCopyRegex(final Object copy) {
         boolean match = false;
-        if (this.classExcludeRegexPatternList.isEmpty()) {
-            if(LOG.isDebugEnabled()) {
-                LOG.warn("classExcludeRegexPatternList is empty, deprecated classExcludeRegexList is used");
-            }
+        if (this.classExcludeRegexPatternList.isEmpty() && this.classExcludeRegexList!=null && !this.classExcludeRegexList.isEmpty()) {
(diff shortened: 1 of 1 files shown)


oleksiimiroshnyk:
>>Display log only if debug is enabled

this most probably means that that log will not be shown. Maybe just delete it ? Idea of that log is to let people know that currently it still works but better do a change.
With the proposed fix I doubt that someone will do any actions...

oleksiimiroshnyk:
>>Display log only if debug is enabled
this most probably means that that log will not be shown. Maybe just delete it ? Idea of that log is to let people know that currently it still works but better do a change.
With the proposed fix I doubt that someone will do any actions...

oleksiimiroshnyk (approved):

riteshadhikari17:
It will not display the log but if you set that class to debug it will. For example in logback setting if you add this:
```
<logger name="org.broadleafcommerce.common.copy.MultiTenantCopier" level="DEBUG"/>
```

riteshadhikari17:
Also it's showing up when duplicating product or when creating new catalog based on other catalogs. So does that mean we are actively using the List in framework? Do we need to make changes in framework to use the pattern?

oleksiimiroshnyk:
> Also it's showing up when duplicating product or when creating new catalog based on other catalogs. So does that mean we are actively using the List in framework? Do we need to make changes in framework to use the pattern?

Good question, quick search didn't find any usage, so I suspect it just evaluates condition to true 
        if (this.classExcludeRegexPatternList.isEmpty()) 
maybe it should be re-written to something like that
if (this.classExcludeRegexPatternList.isEmpty() && this.classExcludeRegexList!=null && !this.classExcludeRegexList.isEmpty())

oleksiimiroshnyk (approved):

PR #2668, opened 2022-04-22
Fix to validate required properties first and not allow rules persistent providers to save entity without required fields

oleksiimiroshnyk:
- Fix to validate required properties first and not allow rules persistent providers to save entity without required fields

Fixes: BroadleafCommerce/QA#4688


StanislavFedorov (approved):
For me works fine. Just one note.
I think we can use this
`entity.isValidationFailure()`

instead of this
`entity.getPropertyValidationErrors() != null && entity.getPropertyValidationErrors().isEmpty()`

StanislavFedorov (approved):

StanislavFedorov (approved):

danielcolgrove:
This fix looks pretty straight forward.  My main concern was how invasive the fix may have been.  If this tests our okay, I'm good with the change.

riteshadhikari17:
@oleksiimiroshnyk I am seeing two error messages. Can you fix this quickly?
image

oleksiimiroshnyk:
@riteshadhikari17 pushed a fix

Issue #1257, opened 2015-02-09
Illegal characters in Solr queries throwing exceptions


brundonsmith:
Use of any Solr query meta characters inside facet values was throwing an exception and causing the query to fail. Facet values need to be scrubbed and special characters escaped.

PR #2671, opened 2022-04-26
Fixed an issue where Order between date filter along with a sort throws error

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/4689

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/listGrid-filter.js
+++ b/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/listGrid-filter.js
@@ -297,10 +297,7 @@ $(document).ready(function() {
 
         $(this).closest('.listgrid-container').find('.mCSB_container').css('top', '0px');
         $(this).closest('.listgrid-container').find('.listgrid-body-wrapper').mCustomScrollbar('update');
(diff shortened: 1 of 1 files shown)


StanislavFedorov:
@danielcolgrove @riteshadhikari17 
I investigated this and the reason for this issue is in this PR 
https://github.com/BroadleafCommerce/BroadleafCommercePrivate/pull/214

Because of this row.:
`var oldParams = BLCAdmin.history.getUrlParameters();`

When we rebuild the URL string with the "Date" type from oldParams - we broke the date format.

At first this logic looks a bit weird to me:
1. We parse URL string and try to grab parameters - we call it "old"
2. Then we try to add the same parameters in the 
`url = getFilteredParams($(this), url);`

So parameters from the url should have higher priority according to this logic. And it should override filters from the "Filter" modal. For me this does not look correct.

When I removed this part of that fix:
```
        var oldParams = BLCAdmin.history.getUrlParameters();
        if (oldParams == null) {
            oldParams = {};
        }
```
For me the described case still works fine
https://github.com/BroadleafCommerce/QA/issues/1257

And our case is also fixed.

danielcolgrove:
If the old use case works and the new use case works, I'm good with the approach.

riteshadhikari17 (approved):

PR #2677, opened 2022-05-16
upgraded spring to 5.2.20 for project 5.2

MarekB01:
upgraded to spring 5.2.20
commented some methods in RuntimeLog4jConfigurer and VelocityMessageCreator also changed scope_global_session to scope_session because scope_global_session is deprecated from spring 5.x

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/email/service/message/VelocityMessageCreator.java
+++ b/common/src/main/java/org/broadleafcommerce/common/email/service/message/VelocityMessageCreator.java
@@ -20,7 +20,7 @@ package org.broadleafcommerce.common.email.service.message;
 import org.apache.velocity.app.VelocityEngine;
 import org.broadleafcommerce.common.email.service.info.EmailInfo;
 import org.springframework.mail.javamail.JavaMailSender;
-import org.springframework.ui.velocity.VelocityEngineUtils;
+//import org.springframework.ui.velocity.VelocityEngineUtils;
 
 import java.util.HashMap;
 import java.util.Map;
@@ -49,7 +49,7 @@ public class VelocityMessageCreator extends MessageCreator {
             if (additionalConfigItems != null) {
                 propsCopy.putAll(additionalConfigItems);
             }
-            return VelocityEngineUtils.mergeTemplateIntoString(velocityEngine, info.getEmailTemplate(), info.getEncoding(), propsCopy);
+            //return VelocityEngineUtils.mergeTemplateIntoString(velocityEngine, info.getEmailTemplate(), info.getEncoding(), propsCopy);
         }
 
         throw new IllegalArgumentException("Property map must be of type HashMap<String, Object>");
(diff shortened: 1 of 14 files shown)


common/src/main/java/org/broadleafcommerce/common/email/service/message/VelocityMessageCreator.java
```diff
@@ -20,7 +20,7 @@
 import org.apache.velocity.app.VelocityEngine;
 import org.broadleafcommerce.common.email.service.info.EmailInfo;
 import org.springframework.mail.javamail.JavaMailSender;
-import org.springframework.ui.velocity.VelocityEngineUtils;
+//import org.springframework.ui.velocity.VelocityEngineUtils;
```
oleksiimiroshnyk, line 23: just remove it if we don't need it

common/src/main/java/org/broadleafcommerce/common/email/service/message/VelocityMessageCreator.java
```diff
@@ -49,7 +49,7 @@ public String buildMessageBody(EmailInfo info, Map<String,Object> props) {
             if (additionalConfigItems != null) {
                 propsCopy.putAll(additionalConfigItems);
             }
-            return VelocityEngineUtils.mergeTemplateIntoString(velocityEngine, info.getEmailTemplate(), info.getEncoding(), propsCopy);
+            //return VelocityEngineUtils.mergeTemplateIntoString(velocityEngine, info.getEmailTemplate(), info.getEncoding(), propsCopy);
```
oleksiimiroshnyk, line 52: think we need to do something else here ? return formatted string ?
check https://stackoverflow.com/questions/14314143/velocityengineutils-has-been-removed-in-spring-3-2-so-what-else-to-use

common/src/main/java/org/broadleafcommerce/common/util/RuntimeLog4jConfigurer.java
```diff
@@ -36,12 +36,12 @@ public String getLog4jConfigLocation() {
...
 
-    public void setLog4jConfigLocation(String log4jConfigLocation) {
-        this.log4jConfigLocation = log4jConfigLocation;
-        try {
-            Log4jConfigurer.initLogging(log4jConfigLocation);
-        } catch (FileNotFoundException e) {
-            throw new RuntimeException(e);
-        }
-    }
+//    public void setLog4jConfigLocation(String log4jConfigLocation) {
```
oleksiimiroshnyk, line 39: need to think do we need this class at all ?
maybe check later versions of framework or looks like it was deprecated in favor of
[org.springframework.util.Log4jConfigurer](https://docs.spring.io/spring-framework/docs/4.2.x/javadoc-api/org/springframework/util/Log4jConfigurer.html)
as of Spring 4.2.1, in favor of Apache Log4j 2 (following Apache's EOL declaration for log4j 1.x)

PR #2678, opened 2022-05-17
Esapi version updated

MarekB01:
https://github.com/BroadleafCommerce/QA/issues/4694
upgraded org.owasp.esapi to version 2.3.0.0 (latest version)

### DIFF
--- a/common/src/main/resources/esapi-java-logging.properties
+++ b/common/src/main/resources/esapi-java-logging.properties
@@ -1,0 +1,23 @@
+###
+# #%L
+# BroadleafCommerce Common Libraries
+# %%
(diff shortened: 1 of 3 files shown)


oleksiimiroshnyk (approved):

riteshadhikari17 (changes requested):
When you upgrade lib, you have to verify that the version is compatible. For that you have to run tests. Always run `mvn clean install` without skipping tests before opening any PR. For the big libs like spring, hibernate, esapi, jquery etc. You should do additional verification by running ReferenceSite.
There are test failures so look at 6.2 and see what was done when upgrading esapi.

riteshadhikari17 (approved):

PR #2684, opened 2022-05-24
Fixed an Error updating Amount Off Item/Order

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/4709


riteshadhikari17:
From next time, remember what I said in scrum. Include more information like what was the problem and which solution you implemented in description.

riteshadhikari17 (approved):

PR #2687, opened 2022-05-25
Fixed an issue where BOGO offers can be created without specifying a qualifier

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/4710

**A Brief Overview**
There is not any sense in defining a BOGO without a qualifier.
The purpose of this type of offer is to give a bonus to a customer that is purchasing some items.
So the qualifiers should be required for BOGO offers.

### DIFF
--- a/admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/persistence/validation/OfferQualifyingCriteriaValidator.java
+++ b/admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/persistence/validation/OfferQualifyingCriteriaValidator.java
@@ -1,0 +1,68 @@
+/*
+ * #%L
+ * BroadleafCommerce Admin Module
+ * %%
+ * Copyright (C) 2009 - 2016 Broadleaf Commerce
+ * %%
+ * Licensed under the Broadleaf Fair Use License Agreement, Version 1.0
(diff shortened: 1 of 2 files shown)


riteshadhikari17:
Its validating but its not right. I added offer, picked bogo then set the rule. I didn't have anything for qualifier and I get this:
image

Expectation: Failure to save and Red border for the field causing error. If possible also look into adding the required indicator(red star next to header) like this. 
image

admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/persistence/validation/OfferQualifyingCriteriaValidator.java
```diff
@@ -0,0 +1,68 @@
+/*
+ * #%L
+ * BroadleafCommerce Admin Module
+ * %%
+ * Copyright (C) 2009 - 2016 Broadleaf Commerce
```
riteshadhikari17: The end year should have been current. Did you copy header instead of generating? It should generate headers when u run mvn install. Anyway change it to 2022.

StanislavFedorov:
Yes, I missed one condition in this validator - I fixed this.
But just one note. This new validator checks if the rule is empty and the validation message is: `This field is required`

image

But on your screenshot, there is another type of validation error - when the rule is not empty, but it is not proper (missing some fields) - `Invalid Rule`

image

riteshadhikari17 (approved):

PR #2695, opened 2022-06-15
Fixed 403 error when page refreshes

MarekB01:
https://github.com/BroadleafCommerce/QA/issues/4723

I changed the logic to use extension handler and changed a validation logic so it can use parent id of product


admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/PersistenceManagerImpl.java
```diff
@@ -421,8 +421,8 @@ protected void postFetchValidation(PersistencePackage persistencePackage, Persis
                         .filter(sectionCrumb -> sectionCrumb.getOriginalSectionIdentifier() != null)
                         .filter(sectionCrumb -> sectionCrumb.getOriginalSectionIdentifier().equals(foreignKey.getManyToField()))
                         .findFirst();
-                    if (property != null && sectionCrumbOptional.isPresent()
-                        && !property.getValue().equals(sectionCrumbOptional.get().getSectionId())) {
+                    property.setValue(sectionCrumbOptional.get().getSectionId());
```
danielcolgrove: You can't change the property.  The solution should be similar to https://github.com/BroadleafCommerce/Enterprise/pull/1452 in that we need to take the property id and find its parent.  The issue here is that this code is in framework and the logic we need is in Enterprise.  So we will likely need an ExtensionHandler to call into Enterprise to get the parent id.

danielcolgrove (changes requested):

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/handler/ParentIdServiceExtensionManager.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package org.broadleafcommerce.openadmin.handler;
+
+import org.broadleafcommerce.common.extension.ExtensionManager;
+import org.springframework.stereotype.Service;
+
+@Service("blParentIdServiceExtensionManager")
+public class ParentIdServiceExtensionManager extends ExtensionManager<BasicParentIdServiceExtensionHandler> {
+
+    public ParentIdServiceExtensionManager(){super(BasicParentIdServiceExtensionHandler.class);}
```
KatrukOV: use code formatting

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/PersistenceManagerImpl.java
```diff
@@ -422,8 +429,16 @@ protected void postFetchValidation(PersistencePackage persistencePackage, Persis
                         .filter(sectionCrumb -> sectionCrumb.getOriginalSectionIdentifier().equals(foreignKey.getManyToField()))
                         .findFirst();
                     if (property != null && sectionCrumbOptional.isPresent()
-                        && !property.getValue().equals(sectionCrumbOptional.get().getSectionId())) {
-                        throw new SecurityServiceException("Post fetch validation: Access denied");
+                            && !property.getValue().equals(sectionCrumbOptional.get().getSectionId())) {
+                        ExtensionResultHolder<String> extensionResultHolder = new ExtensionResultHolder<String>();
+                        if (extensionManager.getProxy().checkParentId(property.getValue(),foreignKey.getForeignKeyClass(), extensionResultHolder) != ExtensionResultStatusType.NOT_HANDLED ){
+                           String parentID = extensionResultHolder.getResult();
```
KatrukOV: AbstractParentIdServiceExtensionHandler.checkParentId() always return ExtensionResultStatusType.NOT_HANDLED;

Is it correct?

riteshadhikari17 (approved):

danielcolgrove (approved):

PR #2696, opened 2022-06-15
Renamed property to differentiate by groupId and updated dependencies from sun.xml which now resolves txw2 properly to 2.3.1.

riteshadhikari17:
**A Brief Overview**
Renamed property to differentiate by groupId. There are no new versions from javax after 2.3.1. So changed one of the version name to `javax` and set it to 2.3.1. Similarly changed another version name to `sunxml` and set to 3.0.2 which was released in Aug 2021. There are newer [milestones](https://mvnrepository.com/artifact/com.sun.xml.ws/jaxws-rt) but decided to use release. 

This resolves txw2 properly to 2.3.1.

**Link to QA**
https://github.com/BroadleafCommerce/QA/issues/4726

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,8 +72,8 @@
         <closure.compiler.version>v20180506</closure.compiler.version>
         <hibernate.version>5.4.33.Final</hibernate.version>
         <hibernate-validator.version>6.1.7.Final</hibernate-validator.version>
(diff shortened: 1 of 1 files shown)


riteshadhikari17:
image

danielcolgrove (approved):

PR #2697, opened 2022-06-15
QA-4728-read-customer-by-email-case-insensitive - modify read customer by email query to be case insensitive

agustinlacombadev:
Modify read customer by email query to be case insensitive

[issue 4728 ](https://github.com/BroadleafCommerce/QA/issues/4728)


oleksiimiroshnyk (approved):
Looks fine, however IIRC with such sql this query will not use index, so in the case of a lot of customers this might be a performance gap. Need to doublecheck but I think I remember correctly.
As alternative solution I would think of doing upper in Java, and for existing clients in migration notes ask to exec sql to make all emails upper-case...

PR #2698, opened 2022-06-21
Deleted weave logic for default sku

MarekB01:
I have deleted weave logic for using default SKU in inventory for version 6.3

https://github.com/BroadleafCommerce/QA/issues/4686


danielcolgrove (approved):

core/broadleaf-framework/src/main/resources/bl-framework-applicationContext.xml
```diff
@@ -257,4 +257,39 @@
         <property name="collectionRef" value="blMergedCacheConfigLocations-framework"/>
         <property name="targetRef" value="blMergedCacheConfigLocations"/>
     </bean>
+
```
riteshadhikari17: I think you need to remove these. There is error `Caused by: org.springframework.beans.factory.NoSuchBeanDefinitionException: No bean named 'blProductInventoryUseDefaultSku' available`.

riteshadhikari17 (approved):

PR #2700, opened 2022-07-07
Client performance solr reindex

StanislavFedorov:
not to merge


oleksiimiroshnyk:
@StanislavFedorov do we still need this PR ? maybe close it if it not needed

PR #2702, opened 2022-07-12
Removed StateImpl bean from context

MarekB01:
https://github.com/BroadleafCommerce/QA/issues/4737

I changed persistence-profile.xml and bl-profile-applicationContext-entity.xml so when program reads context it does not create a StateImpl bean


danielcolgrove:
@riteshadhikari17 @MarekB01 If the StateImpl is gone, we may need data migration scripts.

MarekB01:
> @riteshadhikari17 @MarekB01 If the StateImpl is gone, we may need data migration scripts.
Yeah, I told about it Ritesh. We need to remove BLC_STATE column from database

riteshadhikari17:
@danielcolgrove  It was actually removed in https://github.com/BroadleafCommerce/BroadleafCommerce/pull/2548. These are the references that were missed.

riteshadhikari17:
@MarekB01 Do you mean BLC_STATE table? You will have to remove these https://github.com/BroadleafCommerce/BroadleafCommerce/blob/9cf3f7fab00acc1568cbcee28d5914f44503969d/core/broadleaf-profile/src/main/resources/config/bc/sql/demo/load_code_tables.sql#L284

oleksiimiroshnyk (approved):

riteshadhikari17 (changes requested):
Remove the insert statements from load_code_tables as well

riteshadhikari17 (approved):

PR #2708, opened 2022-08-02
- For org packages use "Fair Use" license not commercial

oleksiimiroshnyk:
- For org packages use "Fair Use" license not commercial
BroadleafCommerce/Client-TuesdayMorning#1623


riteshadhikari17:
Even if this works, we should ask Brad if it's ok to remove Commercial license before merging it.

Jerry77oz (approved):
I checked with Brad and he confirmed that we are good with removing the commercial license from here.

PR #2710, opened 2022-08-11
Order should be read according to its site disc

oleksiimiroshnyk:
BroadleafCommerce/QA#4754
Change from find by id to query so all hibernate filters will be applied correctly, as find by id ignore any filters

@riteshadhikari17 This may need special release notes.  Not sure if any clients use findByID across sites. (@release-notes)

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/dao/OrderDaoImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/dao/OrderDaoImpl.java
@@ -72,7 +72,9 @@ public class OrderDaoImpl implements OrderDao {
 
     @Override
     public Order readOrderById(final Long orderId) {
-        return em.find(OrderImpl.class, orderId);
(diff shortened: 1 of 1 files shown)


core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/dao/OrderDaoImpl.java
```diff
@@ -72,7 +72,9 @@ public class OrderDaoImpl implements OrderDao {
 
     @Override
     public Order readOrderById(final Long orderId) {
-        return em.find(OrderImpl.class, orderId);
+        TypedQuery<Order> query = em.createQuery("SELECT o FROM OrderImpl o WHERE o.id=?1", Order.class);
```
StanislavFedorov: My note is only about syntax. Maybe better to use something like this?

>TypedQuery<Order> query = em.createQuery("SELECT o FROM OrderImpl o WHERE o.id=:orderId", Order.class);
        query.setParameter("orderId", orderId);

StanislavFedorov (approved):

danielcolgrove (approved):

danielcolgrove:
Does the same need to be done to Customer?

PR #2714, opened 2022-08-22
Checkout is not active when product options are not chosen

MarekB01:
https://github.com/BroadleafCommerce/QA/issues/4765

Added new method to Order, so it can reference cart and render it in html


core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/domain/NullOrderImpl.java
```diff
@@ -425,4 +425,9 @@ public <G extends Order> CreateResponse<G> createOrRetrieveCopyInstance(MultiTen
     public Long getBroadleafAccountId() {
         return null;
     }
+
+    @Override
+    public Boolean hasValidationErrors() {
+        return null;
```
oleksiimiroshnyk: do we need here null of false ?

riteshadhikari17 (approved):

PR #2718, opened 2022-08-30
Fixed issue with translation of product options

MarekB01:
https://github.com/BroadleafCommerce/QA/issues/4771

Added some logic which compares option value to key and translates product


core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/OrderItemServiceImpl.java
```diff
@@ -20,12 +20,7 @@
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.collections.MapUtils;
 import org.apache.commons.lang3.builder.CompareToBuilder;
-import org.broadleafcommerce.core.catalog.domain.Category;
-import org.broadleafcommerce.core.catalog.domain.Product;
-import org.broadleafcommerce.core.catalog.domain.ProductBundle;
-import org.broadleafcommerce.core.catalog.domain.ProductOption;
-import org.broadleafcommerce.core.catalog.domain.Sku;
-import org.broadleafcommerce.core.catalog.domain.SkuBundleItem;
+import org.broadleafcommerce.core.catalog.domain.*;
```
danielcolgrove: We don't want to use wildcard imports.

danielcolgrove (changes requested):

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/OrderItemServiceImpl.java
```diff
@@ -125,6 +120,13 @@ protected void populateProductOptionAttributes(OrderItem item, Map<String, Strin
                 if(value == null){
                     continue;
                 }
+                List<ProductOptionValue> optionValues = ((DiscreteOrderItem) item).getSku().getProductOptionValues();
```
oleksiimiroshnyk: I think we can extract this to some method, like String getTranslatedValue(String key) ?

Also getProductOptionValues looks like deprecated let's don't use deprecated in new code ?

danielcolgrove (approved):

oleksiimiroshnyk (approved):

oleksiimiroshnyk (changes requested):
there is a discussion going on in helpscout if this is reasonable or not to do this change.
So let;'s wait and don't merge till then

PR #2722, opened 2022-09-07
Fixed readOrderById when it couldn't find the order

KatrukOV:
**A Brief Overview**
In OrderDaoImpl#readOrderById, if we can't find an order by Id and get NoResultException, we return null;
This allows integration tests to pass as they wait for null when there is no order in the database.

Link to QA [4754](https://github.com/BroadleafCommerce/QA/issues/4754)

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/dao/OrderDaoImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/dao/OrderDaoImpl.java
@@ -90,9 +90,7 @@ public class OrderDaoImpl implements OrderDao {
         try {
             order = query.getSingleResult();
         } catch (NoResultException e) {
-            if (LOG.isWarnEnabled()) {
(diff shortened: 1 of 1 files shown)


oleksiimiroshnyk (approved):

danielcolgrove (approved):

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/dao/OrderDaoImpl.java
```diff
@@ -78,7 +90,9 @@ public Order readOrderById(final Long orderId) {
         try {
             order = query.getSingleResult();
         } catch (NoResultException e) {
-            LOG.warn(String.format("Could not find order by ID %s", orderId));
+            if (LOG.isWarnEnabled()) {
```
danielcolgrove: @KatrukOV This check is not needed.  The LOG.warn does that internally.  Doing a check with isWarnEnabled is good if you have a block of code that gets executed ahead of the logging taking place.

PR #2727, opened 2022-09-16
Upgraded some libraries for 6.3

MarekB01:
Upgraded Spring boot to 2.7.3 and Hibernate to 5.6.11

Link to QA https://github.com/BroadleafCommerce/QA/issues/4786

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/dao/provider/metadata/MapFieldsFieldMetadataProvider.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/dao/provider/metadata/MapFieldsFieldMetadataProvider.java
@@ -39,7 +39,6 @@ import org.hibernate.internal.TypeLocatorImpl;
 import org.hibernate.type.Type;
 import org.hibernate.type.TypeFactory;
 import org.hibernate.type.TypeResolver;
-import org.hibernate.type.internal.TypeConfigurationRegistry;
 import org.hibernate.type.spi.TypeConfiguration;
 import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
@@ -171,7 +170,6 @@ public class MapFieldsFieldMetadataProvider extends DefaultFieldMetadataProvider
                         ((BasicFieldMetadata) entry.getValue()).getExplicitFieldType(),
                         myType.getReturnedClass(), addMetadataFromFieldTypeRequest.getDynamicEntityDao()), metadata);
 
-                TypeConfigurationRegistry.INSTANCE.deregisterTypeConfiguration(typeConfiguration);
             }
         }
         return MetadataProviderResponse.HANDLED;
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -452,5 +452,9 @@
             <groupId>com.googlecode.htmlcompressor</groupId>
             <artifactId>htmlcompressor</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.javassist</groupId>
+            <artifactId>javassist</artifactId>
+        </dependency>
     </dependencies>
 </project>
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -452,9 +452,5 @@
             <groupId>com.googlecode.htmlcompressor</groupId>
             <artifactId>htmlcompressor</artifactId>
         </dependency>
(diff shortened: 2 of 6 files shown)


oleksiimiroshnyk (approved):

common/pom.xml
```diff
@@ -452,5 +452,9 @@
             <groupId>com.googlecode.htmlcompressor</groupId>
             <artifactId>htmlcompressor</artifactId>
         </dependency>
+        <dependency>
```
oleksiimiroshnyk: Why did you add new dependency?

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/dao/provider/metadata/MapFieldsFieldMetadataProvider.java
```diff
@@ -171,7 +170,6 @@ public MetadataProviderResponse addMetadataFromFieldType(AddMetadataFromFieldTyp
                         ((BasicFieldMetadata) entry.getValue()).getExplicitFieldType(),
                         myType.getReturnedClass(), addMetadataFromFieldTypeRequest.getDynamicEntityDao()), metadata);
 
-                TypeConfigurationRegistry.INSTANCE.deregisterTypeConfiguration(typeConfiguration);
```
oleksiimiroshnyk, line 174: is this class missing ? or why did you delete it ?

MarekB01: It is missing, because Hibernate was upgraded to 5.6.11

core/broadleaf-profile/src/main/java/org/broadleafcommerce/profile/core/domain/Country.java
```diff
@@ -33,6 +33,7 @@
  *
  * @author Elbert Bautista (elbertbautista)
  */
+@Deprecated
```
riteshadhikari17: @oleksiimiroshnyk Shouldn't we remove this instead of deprecating? I thought [this ticket](https://github.com/BroadleafCommerce/QA/issues/4509) was for removing. Its 6.3 so its better remove it and in Migration notes mention to use `org.broadleafcommerce.common.i18n.domain.ISOCountryImpl`

riteshadhikari17: @MarekB01 I wasn't talking about removing deprecation tag. I was asking if we should remove these `Country` and `CountryImpl` classes. 
BTW in future, remember to mention the class that should be used along with deprecation tag.

core/broadleaf-profile/src/main/resources/bl-profile-applicationContext-entity.xml
```diff
@@ -23,9 +23,9 @@
            
     <!-- Entity mappings - override for extensibility -->
     <bean id="org.broadleafcommerce.profile.core.domain.Address" class="org.broadleafcommerce.profile.core.domain.AddressImpl" scope="prototype">
-        <property name="country">
-            <bean class="org.broadleafcommerce.profile.core.domain.CountryImpl">
-                <property name="abbreviation" value="US"/>
+        <property name="isoCountryAlpha2">
```
oleksiimiroshnyk, line 26: I think this can be reverter ? as we will not delete CountryImpl

core/broadleaf-profile/src/main/java/org/broadleafcommerce/profile/core/domain/Country.java
oleksiimiroshnyk: @riteshadhikari17 after investigation we decided to not remove CountryImpl, looks like its main purpose is to be "supported country dictionary" while IsoCountry is a global dictionary.
So for now lets leave as it is. 

If we want we can create a separate task to get rid of CountryImpl completely and move it is goal to, let's say, properties or somewhere else.

core/broadleaf-profile/src/main/resources/bl-profile-applicationContext-entity.xml
MarekB01: It cannot see the property with name "country" I think the field Country was removed with State in AddressImpl. I think in this case it would be better to leave the changed code

PR #2730, opened 2022-09-27
If we use default URL, url field becomes read-only

MarekB01:
https://github.com/BroadleafCommerce/QA/issues/4792

Added logic to check if override url is true or false, and if we set default url, url field becomes read-only

### DIFF
--- a/admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/web/controller/entity/AdminProductController.java
+++ b/admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/web/controller/entity/AdminProductController.java
@@ -123,6 +123,21 @@ public class AdminProductController extends AdminBasicEntityController {
         }
     }
 
+
+    @Override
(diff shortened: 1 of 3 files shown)


danielcolgrove:
@MarekB01 The UI looks better but we broke functionality.  Once the product is saved the automatic updates of the URL should no longer take place.

Use Case: New Product
1) Create new product
2) Add name (notice the URL changes)
3) Add category (notice the URL changes)
4) Change name (again, the URL will get updated)
5) Save
6) Change name (now that we've saved, the name will not change - the use now has to change it manually)

Use Case: Existing Product
1) Edit an existing product
2) Change Name (the URL will not change since it was a previously saved product - the use has to enable URL overrides and then manually change)

Once the product is saved, the URL can only be updated manually by changing the "Override URL" to True and then manually updating the URL.  This is because once the URL is established there could be searches, links, etc with this URL and if it is changed things can break.

riteshadhikari17 (approved):

PR #2738, opened 2022-10-05
Removed Spring mobile dependency

MarekB01:
https://github.com/BroadleafCommerce/QA/issues/4735

Removed dependency spring-mobile, also refactored some logic which used it


pom.xml
```diff
@@ -50,7 +50,6 @@
         <spring.version>5.3.20</spring.version>
         <spring.security.version>5.6.6</spring.security.version>
         <spring.boot.version>2.6.7</spring.boot.version>
-        <spring.mobile.version>2.0.0.M3</spring.mobile.version>
```
riteshadhikari17, line 53: We also want to delete `spring-milestones` repo above line#42 to line#45

riteshadhikari17:
Also there is conflict in some class.

riteshadhikari17 (approved):

PR #2741, opened 2022-10-19
Added new exclusions for the SandBoxClassTransformer from the new Java versions

troian88:
Added new exclusions for the SandBoxClassTransformer from the new Java versions

Fixes: BroadleafCommerce/QA#4818


common/src/main/resources/bl-common-applicationContext.xml
```diff
@@ -61,6 +61,8 @@
                             <value>org.broadleafcommerce.openadmin.web.compatibility.JSCompatibilityRequestWrapper</value>
                             <value>org\.hibernate.*</value>
                             <value>org\.quartz.*</value>
+                            <value>org\.terracotta.*</value>
+                            <value>java.*</value>
```
oleksiimiroshnyk, line 65: guess here should be also \.* ?

oleksiimiroshnyk (approved):

riteshadhikari17 (approved):

PR #2745, opened 2022-10-26
Changed deprecated hibernate indexes to JPA indexes

MarekB01:
https://github.com/BroadleafCommerce/QA/issues/4816

Updated indexes, so now we use JPA indexes instead of deprecated Hibernate indexes


oleksiimiroshnyk (approved):

oleksiimiroshnyk:
work will be in https://github.com/BroadleafCommerce/BroadleafCommerce/pull/2833

PR #2746, opened 2022-11-02
Upgraded HSQL version to 2.7.1

MarekB01:
https://github.com/BroadleafCommerce/QA/issues/4813

Updated hsql to version 2.7.1 Also had to add classifier parameter to hsql dependency, because by default it needs JDK-11

### DIFF
--- a/admin/broadleaf-open-admin-platform/pom.xml
+++ b/admin/broadleaf-open-admin-platform/pom.xml
@@ -161,12 +161,6 @@
             <artifactId>commons-pool</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.hsqldb</groupId>
-            <artifactId>hsqldb</artifactId>
-            <version>2.7.1</version>
-            <classifier>jdk8</classifier>
-        </dependency>
-        <dependency>
             <groupId>commons-fileupload</groupId>
             <artifactId>commons-fileupload</artifactId>
         </dependency>
--- a/admin/broadleaf-open-admin-platform/pom.xml
+++ b/admin/broadleaf-open-admin-platform/pom.xml
@@ -163,6 +163,8 @@
         <dependency>
             <groupId>org.hsqldb</groupId>
             <artifactId>hsqldb</artifactId>
+            <version>2.7.1</version>
+            <classifier>jdk8</classifier>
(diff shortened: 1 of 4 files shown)


oleksiimiroshnyk (approved):

admin/broadleaf-open-admin-platform/pom.xml
```diff
@@ -163,6 +163,8 @@
         <dependency>
             <groupId>org.hsqldb</groupId>
             <artifactId>hsqldb</artifactId>
+            <version>2.7.1</version>
```
riteshadhikari17: No need to define version on each of the POMs. It should come from parent POM. Not sure about the classifier so maybe test without it first.

common/pom.xml
```diff
@@ -436,6 +436,8 @@
         <dependency>
             <groupId>org.hsqldb</groupId>
             <artifactId>hsqldb</artifactId>
+            <version>2.7.1</version>
```
riteshadhikari17: Same as above

integration/pom.xml
```diff
@@ -214,6 +214,8 @@
         <dependency>
             <groupId>org.hsqldb</groupId>
             <artifactId>hsqldb</artifactId>
+            <version>2.7.1</version>
```
riteshadhikari17: same as above

riteshadhikari17 (changes requested):

danielcolgrove:
The classifier should be in the project pom (not the parent pom).  Only the version would be in the parent pom.  The project pom can decide if it needs a classifier which we should use classifier as needed.

riteshadhikari17:
@danielcolgrove Classifier is needed for us because we use jdk8 when building. These are needed for ITs only. What if we add test scope to all ? Like this
```
  <dependency>
      <groupId>org.hsqldb</groupId>
      <artifactId>hsqldb</artifactId>
      <classifier>jdk8</classifier>
      <scope>test</scope>
  </dependency>
```

danielcolgrove:
Using the scope should be fine when the scope is needed - for "test" is a good example of when you would want it.

riteshadhikari17 (approved):

PR #2747, opened 2022-11-08
Replaced the reference to Country which is removed in 6.3 with ISOCountry

MarekB01:
https://github.com/BroadleafCommerce/QA/issues/4509

changed applicationContext-entity file not ot use country property of AddressImpl as it was removed


riteshadhikari17:
This PR has nothing to do with lib upgrade Like the title says. Say something like "replaced the reference to Country with ISOCountry" Then in description you can mention that it was removed as part of this https://github.com/BroadleafCommerce/QA/issues/4509

MarekB01:
> This PR has nothing to do with lib upgrade Like the title says. Say something like "replaced the reference to Country with ISOCountry" Then in description you can mention that it was removed as part of this [BroadleafCommerce/QA#4509](https://github.com/BroadleafCommerce/QA/issues/4509)

I plan to make hibernate and spring boot update here, but first we need to merge changes in OMS module

riteshadhikari17:
Ok I'll merge this change so open separate PR for hibernate and other lib upgrade.

riteshadhikari17 (approved):

PR #2748, opened 2022-11-09
Improve exception-handling for the CSRF verification flow

troian88:
Improve exception handling for the CSRF verification

Fixes: BroadleafCommerce/QA#4843

**A Brief Overview**
Return 403 instead of 500

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/security/service/ExploitProtectionService.java
+++ b/common/src/main/java/org/broadleafcommerce/common/security/service/ExploitProtectionService.java
@@ -58,9 +58,9 @@ public interface ExploitProtectionService {
      * @param passedToken The csrf token that was passed in the request
      * @throws SecurityServiceException
      */
-    public void compareToken(String passedToken) throws SecurityServiceException;
+    public void compareToken(String passedToken) throws ServiceException;
 
-    public String getCSRFToken() throws SecurityServiceException;
+    public String getCSRFToken() throws ServiceException;
 
     public String getCsrfTokenParameter();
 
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/broadleafcommerce/common/security/service/ExploitProtectionService.java
```diff
@@ -55,11 +56,11 @@ public interface ExploitProtectionService {
...
      *
      * @param passedToken The csrf token that was passed in the request
-     * @throws ServiceException
+     * @throws SecurityServiceException
      */
-    public void compareToken(String passedToken) throws ServiceException;
+    public void compareToken(String passedToken) throws SecurityServiceException;
 
-    public String getCSRFToken() throws ServiceException;
+    public String getCSRFToken() throws SecurityServiceException;
```
oleksiimiroshnyk: I would not change this, only compare token should generate SecurityException

common/src/main/java/org/broadleafcommerce/common/security/service/ExploitProtectionServiceImpl.java
```diff
@@ -138,18 +138,18 @@ protected boolean hasHTMLTags(String text){
...
+                throw new SecurityServiceException("XSRF token mismatch (" + passedToken + "). Session may be expired.");
             } else {
                 LOG.debug("Validated CSRF token");
             }
         }
     }
 
     @Override
-    public String getCSRFToken() throws ServiceException {
+    public String getCSRFToken() throws SecurityServiceException {
```
oleksiimiroshnyk: let;s revert

oleksiimiroshnyk (approved):

PR #2750, opened 2022-11-11
Remove redundant weave as it was included in main code

oleksiimiroshnyk:
- Remove redundant weave as it was included in main code

Fixes: BroadleafCommerce/QA#4841


riteshadhikari17:
There is conflict and I tried merging and rebase but it is so far behind develop branch. Also there is issue on UI.
image

oleksiimiroshnyk:
@riteshadhikari17 I pushed some changes should be ok now...seems at some point of time name of property was changed but tooltip etc was referring old name

riteshadhikari17:
It is breaking some tests now. 
```
[ERROR] Errors: 
[ERROR]   ValidateAvailabilityActivitySpec.If the order item is non-null, and there is a BundleOrderItem, then a Sku from that BundleOrderItem will be tested for availability:90 » NullPointer
[ERROR]   ValidateAvailabilityActivitySpec.If the order item is non-null, and there is a DiscreteOrderItem, then a Sku from that DiscreteOrderItem will be tested for availability:72 » NullPointer
[INFO] 
[ERROR] Tests run: 199, Failures: 1, Errors: 2, Skipped: 0
```

riteshadhikari17 (approved):

PR #2751, opened 2022-11-11
make sure that if value is string but metadata is broadleaf enumeration final value will be a friendly type of broadleaf enumeration

oleksiimiroshnyk:
- make sure that if value is string but metadata is broadleaf enumeration final value will be a friendly type of broadleaf enumeration

Fixes: BroadleafCommerce/QA#4846


KatrukOV (approved):
We should improve this (add a method to the interface and not use reflection) in 6.3

PR #2752, opened 2022-11-14
Updated hibernate to version 5.6.14

MarekB01:
https://github.com/BroadleafCommerce/QA/issues/4786

Updated hibernate to 5.6.14, Also removed deprecated TypeConfigurationRegistry class and added javassist dependency in pom because it was removed in Hibernate 5.6 and is not included in it

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -442,5 +442,10 @@
             <groupId>com.googlecode.htmlcompressor</groupId>
             <artifactId>htmlcompressor</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.javassist</groupId>
+            <artifactId>javassist</artifactId>
+            <version>3.22.0-GA</version>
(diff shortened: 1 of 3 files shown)


common/pom.xml
```diff
@@ -442,5 +442,10 @@
             <groupId>com.googlecode.htmlcompressor</groupId>
             <artifactId>htmlcompressor</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.javassist</groupId>
+            <artifactId>javassist</artifactId>
+            <version>3.22.0-GA</version>
```
oleksiimiroshnyk: i'm not sure if we want to specify version here.
I think we usually have somewhere in "root" poms dependencyManagement section where we declare dependencies and its versions... can you please check ?

oleksiimiroshnyk (approved):

PR #2756, opened 2022-11-28
Got rid of deprecated processors in Core module

MarekB01:
https://github.com/BroadleafCommerce/QA/issues/4832

We got rid of deprecated processors in CommonPresentation module. There were a lot of classes which were implementing deprecated logic, so I made some new interfaces as it is mentioned in deprecation docs (@deprecated instead of using this processor you should instead write a custom {@link org.broadleafcommerce.common.web.expression.BraodleafVariableExpression})

### DIFF
--- a/admin/broadleaf-contentmanagement-module/src/main/java/org/broadleafcommerce/cms/web/processor/ContentExpression.java
+++ b/admin/broadleaf-contentmanagement-module/src/main/java/org/broadleafcommerce/cms/web/processor/ContentExpression.java
@@ -1,0 +1,36 @@
+/*-
+ * #%L
+ * BroadleafCommerce CMS Module
+ * %%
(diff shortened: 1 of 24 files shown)


riteshadhikari17:
@MarekB01 Shouldn't this be for 6.3?

MarekB01:
> [BroadleafCommerce/QA#4832](https://github.com/BroadleafCommerce/QA/issues/4832)
> 
> We got rid of deprecated processors in CommonPresentation module. There were a lot of classes which were implementing deprecated logic, so I made some new interfaces as it is mentioned in deprecation docs (@deprecated instead of using this processor you should instead write a custom {@link org.broadleafcommerce.common.web.expression.BraodleafVariableExpression})

Closed because of deprecation of this solution

PR #2764, opened 2022-12-28
Sku with future pricelist data now don't disappear from the site 

MarekB01:
https://github.com/BroadleafCommerce/QA/issues/4869

Removed a logic which checks if Sku has dynamics active dates, as it was blocking the update of Sku price with new pricelist data applied

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/catalog/domain/SkuImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/catalog/domain/SkuImpl.java
@@ -53,7 +53,6 @@ import org.broadleafcommerce.common.presentation.client.VisibilityEnum;
 import org.broadleafcommerce.common.util.DateUtil;
 import org.broadleafcommerce.common.web.BroadleafRequestContext;
 import org.broadleafcommerce.core.catalog.service.dynamic.DynamicSkuPrices;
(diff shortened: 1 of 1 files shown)


core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/catalog/domain/SkuImpl.java
```diff
@@ -821,19 +820,10 @@ public void setAvailable(Boolean available) {
 
     @Override
     public Date getActiveStartDate() {
```
oleksiimiroshnyk, line 821: And what about active end date ? I think this should also changed

oleksiimiroshnyk (approved):

riteshadhikari17:
@MarekB01  Did you test this? For me its not disappearing but its not honoring the start date. I have set the date in future but it shows up right away.

1) Create a pricelist for Logged In customers. There is a segment for Logged in customer, use that for Rule. Set priority as 1 and currency as USD and parent as USD. Now this PL should apply for any logged in customer.
2) Add PL data to any product. Lets say set $4.45 for Hopping sauce and put in some future date for startDate:
image
3) Check site as guest which should say $4.99. Log in as any cust and it will say $4.45.

Expected: 4.45 is set for future so it should not display that price.

PR #2765, opened 2023-01-04
Added cleanup method for the Email Tracking DAO

troian88:
Added cleanup method for Email Tracking DAO

**Additional context**
https://github.com/BroadleafCommerce/QA/issues/4865

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/email/dao/EmailReportingDaoImpl.java
+++ b/common/src/main/java/org/broadleafcommerce/common/email/dao/EmailReportingDaoImpl.java
@@ -20,14 +20,22 @@ package org.broadleafcommerce.common.email.dao;
 import org.broadleafcommerce.common.email.domain.EmailTarget;
 import org.broadleafcommerce.common.email.domain.EmailTracking;
 import org.broadleafcommerce.common.email.domain.EmailTrackingClicks;
+import org.broadleafcommerce.common.email.domain.EmailTrackingImpl;
 import org.broadleafcommerce.common.email.domain.EmailTrackingOpens;
 import org.broadleafcommerce.common.persistence.EntityConfiguration;
 import org.broadleafcommerce.common.time.SystemTime;
+import org.hibernate.jpa.QueryHints;
 import org.springframework.stereotype.Repository;
 
 import javax.annotation.Resource;
 import javax.persistence.EntityManager;
 import javax.persistence.PersistenceContext;
+import javax.persistence.Query;
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Root;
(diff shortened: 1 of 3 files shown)


common/src/main/java/org/broadleafcommerce/common/email/dao/EmailReportingDaoImpl.java
```diff
@@ -61,6 +69,25 @@ public EmailTarget createTarget() {
...
+        CriteriaBuilder builder = em.getCriteriaBuilder();
+        CriteriaQuery<EmailTracking> criteria = builder.createQuery(EmailTracking.class);
+        Root<EmailTrackingImpl> EmailTrackingRoot = criteria.from(EmailTrackingImpl.class);
+
+        criteria.select(EmailTrackingRoot);
+
+        Query query = em.createQuery(criteria);
+        query.setHint(QueryHints.HINT_CACHEABLE, true);
+
+        for (Object emailTracking : query.getResultList()) {
```
oleksiimiroshnyk: I think potentially we can have large number of records here, so this is not optimal
I think we can do this via JPQL ?
https://vladmihalcea.com/bulk-update-delete-jpa-hibernate/

oleksiimiroshnyk (changes requested):

common/src/main/java/org/broadleafcommerce/common/email/dao/EmailReportingDaoImpl.java
troian88: updated

common/src/main/java/org/broadleafcommerce/common/email/dao/EmailReportingDaoImpl.java
```diff
@@ -61,6 +66,14 @@ public EmailTarget createTarget() {
         return target;
     }
 
+    @Override
+    public void clearAllRecordsOlderThan(LocalDateTime date) {
+        int deleteCount = em.createQuery("DELETE FROM BLC_EMAIL_TRACKING WHERE DATE_SENT <= :date")
+            .setParameter("date", Timestamp.valueOf(date))
+            .executeUpdate();
+        LOG.debug("clearAllRecordsOlderThan: " + deleteCount + " items was removed.");
```
oleksiimiroshnyk: I would probably make this more informative: deleted X records from BLC_EMAIL_TRACKING with date_sent less than Y ?
and maybe level info - we don't expect it to be executed too often, but can be debug, not sure.

oleksiimiroshnyk (approved):

danielcolgrove:
@troian88 @oleksiimiroshnyk The blc_email_tracking table does not have an index on date_sent...to prevent a table scan should we add one?

danielcolgrove (approved):

riteshadhikari17 (approved):

PR #2771, opened 2023-01-17
Updating spring security version in pom

KatrukOV:
**A Brief Overview**
Updated spring.security.version to 5.8.1

**Link to QA issue**
[Link ](https://github.com/BroadleafCommerce/QA/issues/4823)

**Additional context**
This property is best kept in sync with the version in the broadleaf-module-parent.


oleksiimiroshnyk (approved):

riteshadhikari17:
Just FYI, this is core and it doesn't use module parent or boot starter. So it's necessary to maintain versions in this pom.

PR #2779, opened 2023-02-09
Fixed alignment on filter button

MarekB01:
Fixes: https://github.com/BroadleafCommerce/QA/issues/4888

Added top parameter for filter-info block so now it is on same level as search button

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/css/admin/entityForm.css
+++ b/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/css/admin/entityForm.css
@@ -263,11 +263,8 @@ button:disabled:hover, button.disabled:hover {
 .filter-info {
     float: right;
     text-align: right;
-    top: -5px;
 }
-.filter-info .button-group {
-    margin-top: 6px;
-}
+
 .filter-info > .filter-button {
     margin-top: 6px;
 }
--- a/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/css/admin/entityForm.css
+++ b/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/css/admin/entityForm.css
@@ -263,6 +263,7 @@ button:disabled:hover, button.disabled:hover {
 .filter-info {
     float: right;
     text-align: right;
+    top: -5px;
 }
 .filter-info .button-group {
     margin-top: 6px;


StanislavFedorov (approved):

admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/css/admin/entityForm.css
```diff
@@ -263,6 +263,7 @@ button:disabled:hover, button.disabled:hover {
 .filter-info {
     float: right;
     text-align: right;
+    top: -5px;
```
oleksiimiroshnyk: I think this change will make it "higher" when filter is not applied
I think the issue is here that once filter is applied filter button with "X" button are wrapped into div with style .filter-info .button-group {
    margin-top: 6px;
}
.filter-info > .filter-button {
    margin-top: 6px;
}
that does this 6px offset. So we can remove it at all, or wrap search elements also in some kind of this thing, or just add new style when filter is applied etc if we want to have them in synch

@StanislavFedorov can you take it from there ?

oleksiimiroshnyk (changes requested):

admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/css/admin/entityForm.css
StanislavFedorov: @oleksiimiroshnyk 
I agree that better to remove this
```
.filter-info .button-group {
    margin-top: 6px;
}
```
but before this please @MarekB01 go through the admin UI and check if we have this button-group in other places

MarekB01: @StanislavFedorov Seems like this button-group is only filter related. Also I tried to remove
.filter-info .button-group {
    margin-top: 6px;
}
But it didn't help

StanislavFedorov: @MarekB01 
**Also I tried to remove .. But it didn't help**

Do you mean this css? For me removing exactly this css block works

```
.filter-info .button-group {
    margin-top: 6px;
}
```

MarekB01: @StanislavFedorov now it works for me too

oleksiimiroshnyk (approved):

PR #2800, opened 2023-02-23
Fixed Thymeleaf custom tags usages

troian88:
Update custom expression processors and views

**Link to QA**
https://github.com/BroadleafCommerce/QA/issues/4913

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/templates/layout/partials/footer.html
+++ b/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/templates/layout/partials/footer.html
@@ -7,8 +7,7 @@
     <blc:admin_user>
     <blc:form id="js-error-handling" method="POST" ></blc:form>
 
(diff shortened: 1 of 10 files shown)


admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/templates/layout/partials/footer.html
```diff
@@ -7,8 +7,7 @@
     <blc:admin_user>
```
oleksiimiroshnyk, line 7: guess we can delete this tag as well ?

danielcolgrove:
We definitely want to understand what the intent of the tag was and whether we still need it.  It will also be important that we document what all we are removing.  We need to make sure our migration notes are thorough and have these details.

admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/templates/layout/partials/footer.html
troian88: fixed

oleksiimiroshnyk (approved):

PR #2808, opened 2023-03-14
Fix TypeError error is triggered while resizing the window for the custom listgrid

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/4932

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/listGrid-paginate.js
+++ b/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/listGrid-paginate.js
@@ -766,7 +766,7 @@
             // If we're the only grid on the page, we should stretch to the bottom of the screen if we are not encapsulated
             // inside of an entity-form
             var listGridsCount = BLCAdmin.listGrid.getListGridCount($);
-            if (listGridsCount == 1 && $wrapper.parents('.entity-form').length == 0 &&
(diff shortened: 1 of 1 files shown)


admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/js/admin/components/listGrid-paginate.js
```diff
@@ -766,7 +766,7 @@
             // If we're the only grid on the page, we should stretch to the bottom of the screen if we are not encapsulated
             // inside of an entity-form
             var listGridsCount = BLCAdmin.listGrid.getListGridCount($);
-            if (listGridsCount == 1 && $wrapper.parents('.entity-form').length == 0 &&
+            if (listGridsCount == 1 && $wrapper.length !== 0 && $wrapper.parents('.entity-form').length == 0 &&
```
oleksiimiroshnyk, line 769: well, I think this fix means that grid was not initialized, like in listGrid-paginate there is a initialiaze function that will wrap list grid into wrapper.
I'm not sure if this will affect or not affect anything else. 
I would suggest flywire try to initialize list grid and not do this changes.
At least I don't think we need them in the framework 
@Jerry77oz what do you think ?

Jerry77oz: I agree with @oleksiimiroshnyk, this sounds like it was more of an issue with the Flywire project and since @StanislavFedorov was able to fix this in their project, I think we can bypass this PR. I will close it.

PR #2816, opened 2023-04-04
Added a regex validation for admin password

MarekB01:
https://github.com/BroadleafCommerce/QA/issues/4943

Created AdminUserRegexValidator to set Regex and validate password with it. Also set new properties in common.properties file, so customers could set their own regex and use this logic.


admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/validation/AdminUserRegexValidator.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Override
+    public PropertyValidationResult validate(Entity entity,
+                                             Serializable instance,
+                                             Map<String, FieldMetadata> entityFieldMetadata,
+                                             Map<String, String> validationConfiguration,
+                                             BasicFieldMetadata propertyMetadata,
+                                             String propertyName,
+                                             String value) {
+        String expression = validationConfiguration.get("regularExpression");
```
oleksiimiroshnyk: I think we don't need to get regexp from configuration validationConfiguration.get("regularExpression"); as we expect to come it from properties file
```suggestion

```

admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/validation/AdminUserRegexValidator.java
```diff
@@ -0,0 +1,75 @@
...
+
+        if (value == null) {
```
oleksiimiroshnyk: I think no need to check value, which is password value for null, as we validate through regex. Or if regex validation will fail with NPE just put empty string there

oleksiimiroshnyk (approved):

PR #2835, opened 2023-05-19
Update version in POMs

KatrukOV:
**A Brief Overview**
Updated versions in pom: spring, closure.compiler, jacoco-maven-plugin, groovy; remove commons-logging

Link to QA issue
[QA-4983](https://github.com/BroadleafCommerce/QA/issues/4983)

### DIFF
--- a/integration/pom.xml
+++ b/integration/pom.xml
@@ -202,6 +202,16 @@
             <artifactId>spring-boot-test</artifactId>
             <version>${spring.boot.version}</version>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>jakarta.xml.bind</groupId>
+                    <artifactId>jakarta.xml.bind-api</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>jakarta.annotation</groupId>
+                    <artifactId>jakarta.annotation-api</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>javax.servlet</groupId>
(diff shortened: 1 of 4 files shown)


agustinlacombadev (approved):

integration/pom.xml
```diff
@@ -202,6 +202,16 @@
             <artifactId>spring-boot-test</artifactId>
             <version>${spring.boot.version}</version>
             <scope>test</scope>
+            <exclusions>
```
oleksiimiroshnyk, line 205: looks like this changes are from 6.3 not 6.2 ?

oleksiimiroshnyk (approved):

riteshadhikari17:
~~DO NOT MERGE!! These exclusions are sensitive things. I'll have test this with plugins like 'dependency-check' and 'enforcer' as well as verify versions and make sure it didn't break anything.~~
Also there are new vulnerabilities being reported so will probably have to update other libs too.

EDIT: Updated

integration/pom.xml
riteshadhikari17: @KatrukOV Is this relevant? It seems ok in my smoke test and doesn't seem to cause issues. Have you tested these changes with Refsite? I am ok with merging this if alex is ok with this.

KatrukOV: Yes, this is relevant, we checked it.

oleksiimiroshnyk: I'm ok with this change

riteshadhikari17 (approved):

PR #2851, opened 2023-07-06
Add validation for product and category URI - special characters are not allowed

StanislavFedorov:
**A Brief Overview**
In validation we use the same logic as in the URI generator - we allow only Alphanumeric characters and "-", "_", "/"

https://github.com/BroadleafCommerce/QA/issues/5016


agustinlacombadev (approved):
LGTM

oleksiimiroshnyk (approved):

PR #2854, opened 2023-07-12
Fixed an issue when Searching for catalogs using % in the name throws an error

StanislavFedorov:
**A Brief Overview**
Fixed an issue when Searching for catalogs using % in the name throws an error

**Link to QA issue**
[Link](https://github.com/BroadleafCommerce/QA/issues/5020)

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/controller/AdminAbstractController.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/controller/AdminAbstractController.java
@@ -464,10 +464,10 @@ public abstract class AdminAbstractController extends BroadleafAbstractControlle
                     List<String> values = entry.getValue();
                     List<String> collapsedValues = new ArrayList<>();
                     for (String value : values) {
-                        String decoded = null;
+                        String decoded = value;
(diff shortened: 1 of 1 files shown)


admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/web/controller/AdminAbstractController.java
```diff
@@ -464,10 +464,10 @@ protected FilterAndSortCriteria[] getCriteria(Map<String, List<String>> requestP
                     List<String> values = entry.getValue();
                     List<String> collapsedValues = new ArrayList<>();
                     for (String value : values) {
-                        String decoded = null;
+                        String decoded = value;
                         try {
                             decoded = eps.cleanString(URLDecoder.decode(value, "UTF-8"));
-                        } catch (UnsupportedEncodingException | ServiceException e) {
+                        } catch (Exception e) {
```
oleksiimiroshnyk, line 470: I think we should try to encode % after decoding...so sql will treat % as a regular symbol not special char that means _any_ zero or more

oleksiimiroshnyk (changes requested):

oleksiimiroshnyk (approved):

PR #2857, opened 2023-07-24
Jquery error in roles filter logic

MarekB01:
Fixes: https://github.com/BroadleafCommerce/QA/issues/5026

Fixed Jquery error in role management page on filters


StanislavFedorov:
Is there some way to prevent the Java Script Error without changing addType?
For me this appearance is not the best because when we have a dozens permissions in 1 role - it looks a bit ugly:
image

PR #2858, opened 2023-08-01
Got rid of old spring-social dependency 

MarekB01:
Related to: https://github.com/BroadleafCommerce/QA/issues/4974

Removed spring-social dependency as it became deprecated. Rewrote BroadleafSocialRegisterController to use OAuth2 implementation.


riteshadhikari17:
This is the error in jenkins
```
[ERROR]     'dependencies.dependency.version' for org.springframework.security:spring-security-oauth2-client:jar is missing. @ line 101, column 21
```

FYI Module Parent is for modules. Core does not use MP. Define the version here and close the PR for MP.

MarekB01:
Now it can be merged

core/broadleaf-profile-web/pom.xml
```diff
@@ -81,5 +81,11 @@
             <groupId>org.springframework</groupId>
             <artifactId>spring-oxm</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.security</groupId>
+            <artifactId>spring-security-oauth2-client</artifactId>
+            <version>5.5.3</version>
```
riteshadhikari17: Don't define version here. It should come from root pom.

pom.xml
```diff
@@ -1162,6 +1162,11 @@
                 <artifactId>spring-boot-autoconfigure</artifactId>
                 <version>${spring.boot.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.springframework.security</groupId>
+                <artifactId>spring-security-oauth2-client</artifactId>
+                <version>6.0.2</version>
```
riteshadhikari17: This version needs to sync with spring security. We define it as property above so use it.
`${spring.security.version}`. Also shouldn't you remove the spring-social dependency below?

core/broadleaf-framework-web/pom.xml
```diff
@@ -98,5 +98,9 @@
             <groupId>org.easymock</groupId>
             <artifactId>easymockclassextension</artifactId>
         </dependency>
+        <dependency>
```
oleksiimiroshnyk, line 97: I see you've added a new dependency, but what about deleting old spring social ? Is it still there ?

oleksiimiroshnyk (changes requested):

riteshadhikari17 (approved):

oleksiimiroshnyk (approved):

PR #2860, opened 2023-08-03
Updated deprecated design in @GenericGenerator for Hibernate 6

MarekB01:
Related to: https://github.com/BroadleafCommerce/QA/issues/4999

property strategy was replaced with type:

before: strategy="org.broadleafcommerce.common.persistence.IdOverrideTableGenerator"

after: type= IdOverrideTableGenerator.class


riteshadhikari17 (changes requested):
Your IDE is changing the order of imports when it's optimizing them. Could you please change the layout in settings and revert them?

riteshadhikari17 (approved):

oleksiimiroshnyk (approved):

PR #2871, opened 2023-08-18
Fix to allow empty value in admin regex validator, empty values passes validation

oleksiimiroshnyk:
- Fix to allow empty value in regex validator, empty values passes validation

Fixes: BroadleafCommerce/QA#5036


Jerry77oz (approved):

Jerry77oz:
This has been validated by the client via a hotfix.

PR #2873, opened 2023-08-21
Upgrade of closure-compiler-unshaded

StanislavFedorov:
**A Brief Overview**
Upgrade of closure-compiler-unshaded to version v20230802

https://github.com/BroadleafCommerce/QA/issues/5039

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -420,5 +420,13 @@
             <groupId>com.googlecode.htmlcompressor</groupId>
             <artifactId>htmlcompressor</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protobuf-java</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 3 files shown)


riteshadhikari17:
This can't be upgraded because it needs java upgrade.
```
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/jenkins/agent/workspace/ade-of-closure-compiler-unshaded/common/src/main/java/org/broadleafcommerce/common/resource/service/GoogleClosureJavascriptMinificationServiceImpl.java:[25,36] cannot access com.google.javascript.jscomp.AbstractCommandLineRunner
  bad class file: /home/jenkins/.m2/repository/com/google/javascript/closure-compiler-unshaded/v20230802/closure-compiler-unshaded-v20230802.jar(com/google/javascript/jscomp/AbstractCommandLineRunner.class)
    class file has wrong version 55.0, should be 52.0
    Please remove or make sure it appears in the correct subdirectory of the classpath.
```

StanislavFedorov:
@riteshadhikari17 
hmm. For me locally it worked .. And for customers too. Could you check this variant, please? I have chosen the minimal version that ok with customers' requirements.
If this will not work - then we need to downgrade the closure library

StanislavFedorov:
@riteshadhikari17 (cc: @oleksiimiroshnyk )
I downgraded to v20200830 - this is the latest version for Java 8 and without any **javax.annotation.***
But I did not revert changes in GoogleClosureJavascriptMinificationServiceImpl.java - so customers could update **closure** lib on their side - just change the version

oleksiimiroshnyk (approved):

riteshadhikari17 (approved):

PR #2887, opened 2023-09-19
Make all tests to run during maven build

oleksiimiroshnyk:
- Use junit "vintage engine" as we still have old junit 3/4 tests and in classpath there is junit 5
- Add logback configs for test so it will be found, seems log4j2 is not picked up/selected regardless it is in classpath
- 
Fixes: BroadleafCommerce/QA#5076

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -416,7 +416,6 @@
         <dependency>
             <groupId>org.hsqldb</groupId>
             <artifactId>hsqldb</artifactId>
-            <classifier>debug</classifier>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -104,6 +104,16 @@
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
         </dependency>
+        <!-- seems some lib brings junit 5 in classpath and so surefire selects engine to be junit platform 5
+             while we still have old junit 4(and I think junit3) tests and some of them not annotated with @Test
+             Also seems @Test annotation was moved in junit 5 to another package and Junit4 annotation is ignored
+             as well as "extends TestCase" and "ends on Test". So for now to help surefire to understand that we still
+             want to use Junit4 we need to specify engine that can do that
+        -->
+        <dependency>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
+        </dependency>
         <dependency>
             <groupId>org.mvel</groupId>
             <artifactId>mvel2</artifactId>
@@ -137,6 +147,11 @@
         <dependency>
             <groupId>org.hibernate.orm</groupId>
             <artifactId>hibernate-jcache</artifactId>
+        </dependency>
+        <!-- this is needed for hibernate 6.2, when running tests it will try to transform classes and needs bytecode lib -->
+        <dependency>
+            <groupId>net.bytebuddy</groupId>
+            <artifactId>byte-buddy</artifactId>
         </dependency>
         <dependency>
             <groupId>org.ehcache</groupId>
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -426,7 +426,7 @@
         <dependency>
             <groupId>org.hsqldb</groupId>
             <artifactId>hsqldb</artifactId>
-            <classifier>jdk8</classifier>
+            <classifier>debug</classifier>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/common/src/test/java/org/broadleafcommerce/test/common/rule/MvelHelperTest.java
+++ b/common/src/test/java/org/broadleafcommerce/test/common/rule/MvelHelperTest.java
@@ -40,7 +40,6 @@ import junit.framework.TestCase;
 public class MvelHelperTest extends TestCase {
 
     private static final Log LOG = LogFactory.getLog(MvelHelperTest.class);
-    public static boolean something = false;
 
     /**
      * Test that a blank rule is true.
@@ -146,7 +145,6 @@ public class MvelHelperTest extends TestCase {
                 break;
             }
         }
-        something=false;
         assertEquals("true", output);
     }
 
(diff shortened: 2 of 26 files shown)


riteshadhikari17:
This is the error in jenkins
```
[ERROR] org.broadleafcommerce.test.common.rule.MvelHelperTest.testMvelMethodOverloadFailureCase -- Time elapsed: 0.208 s <<< ERROR!
java.io.IOException: 
Error: Could not find or load main class org.broadleafcommerce.test.common.rule.MvelOverloadFailureReproduction
Caused by: java.lang.ClassNotFoundException: org.broadleafcommerce.test.common.rule.MvelOverloadFailureReproduction
```

riteshadhikari17:
The lines I am talking about are coming from MvelTestUtils#line136. 

By the way I am checking rest of the tests and integration still has issues for 7.0
```
[INFO] --- maven-surefire-plugin:3.1.2:test (default-test) @ integration ---
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
```

This is 6.2
```
[INFO] --- maven-surefire-plugin:2.22.2:test (default-test) @ integration ---
[INFO] Tests run: 167, Failures: 0, Errors: 0, Skipped: 0
```

common/pom.xml
```diff
@@ -401,6 +416,7 @@
         <dependency>
             <groupId>org.hsqldb</groupId>
             <artifactId>hsqldb</artifactId>
+            <classifier>debug</classifier>
```
riteshadhikari17: Is this classifier necessary?

common/src/test/java/org/broadleafcommerce/test/common/rule/MvelHelperTest.java
```diff
@@ -40,6 +40,7 @@
 public class MvelHelperTest extends TestCase {
 
     private static final Log LOG = LogFactory.getLog(MvelHelperTest.class);
+    public static boolean something = false;
```
riteshadhikari17: Looks like added for debugging. If not, maybe a meaningful var name.

common/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,44 @@
...
+  Alternatively, the Commercial License may be replaced with a mutually agreed upon license (the "Custom License")
+  between you and Broadleaf Commerce. You may not use this file except in compliance with the applicable license.
+  #L%
+  -->
+
+<Configuration status="WARN">
+    <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <Pattern>
+                %d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n
```
riteshadhikari17: I have this pattern for all modules. lets stay consistent
`[%-5level] %d{HH:mm:ss.SSS} [%t] %logger{36} - %msg%n`

integration/pom.xml
```diff
@@ -58,7 +71,9 @@
                             <goal>test</goal>
                         </goals>
                         <configuration>
+<!--                            <argLine>-Dmaven.surefire.debug -Xmx1024m ${surefire.argLine}</argLine>-->
```
riteshadhikari17, line 74: Can we delete these commented lines?

integration/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,43 @@
...
+  Alternatively, the Commercial License may be replaced with a mutually agreed upon license (the "Custom License")
+  between you and Broadleaf Commerce. You may not use this file except in compliance with the applicable license.
+  #L%
+  -->
+
+<Configuration>
+    <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <Pattern>
+                %d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n
```
riteshadhikari17: Same as above. pattern `[%-5level] %d{HH:mm:ss.SSS} [%t] %logger{36} - %msg%n`

pom.xml
```diff
@@ -1285,6 +1291,7 @@
                 <groupId>org.hsqldb</groupId>
                 <artifactId>hsqldb</artifactId>
                 <version>${hsqldb.version}</version>
+                <classifier>debug</classifier>
```
riteshadhikari17: same as above.

integration/pom.xml
```diff
@@ -216,11 +235,12 @@
...
             <artifactId>hsqldb</artifactId>
+            <classifier>debug</classifier>
         </dependency>
-        <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-        </dependency>
+        <!-- not adding junit for now, as we have old junit4
+            and spock/gmaven will compile with junit5 support
+            so rely on junit(junit-platform) that comes fomr spring/groovy
```
riteshadhikari17: Typo. change `fomr` to `from`

riteshadhikari17 (changes requested):
For me I see a failure in `CatalogTest` locally (both IDE and shell) but I don't see that being run on jenkins build. Could you check what's going on? Weird thing is there are only 161 tests in older builds so it is picking up new things.
```
[ERROR] Tests run: 161, Failures: 1, Errors: 0, Skipped: 10, Time elapsed: 12.25 s <<< FAILURE! -- in TestSuite
[ERROR] org.broadleafcommerce.core.catalog.service.CatalogTest.testCatalog -- Time elapsed: 0.010 s <<< FAILURE!
java.lang.AssertionError
	at org.broadleafcommerce.core.catalog.service.CatalogTest.testCatalog(CatalogTest.java:132)
```

riteshadhikari17:
image

riteshadhikari17:
It passes if I run the test individually but fails with I run whole module. It looks like it's using previous state because there is 't-shirt' as first category. `OrderTest` fails even if this passes.
image

common/pom.xml
oleksiimiroshnyk: Not really, it helped me to identify the issue with tests/that spring testng integration creates multiple contexts...
Will remove it

integration/pom.xml
oleksiimiroshnyk: I would like to keep them, just in case the next time will need to debug tests that are running through surefire plugin it could save time to find those arguments

oleksiimiroshnyk:
Hm, seems order of execution is different for some reason... added logic to clean up categories. Hope this will not break something else

KatrukOV (approved):

riteshadhikari17 (approved):

riteshadhikari17:
Merging this for now. Will handle local issue separately.

PR #2899, opened 2023-09-29
Removed legacy xxx classes

KatrukOV:
**A Brief Overview**
Removed legacy xxx classes;

**Link to QA issue**
[QA-4923](https://github.com/BroadleafCommerce/QA/issues/4923)


core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/catalog/domain/ProductImpl.java
```diff
@@ -741,21 +742,30 @@ public void setCategory(Category category) {
 
     @Override
     public Map<String, Media> getMedia() {
-        return getDefaultSku().getSkuMedia();
+        return getDefaultSku().getSkuMediaXref()
```
oleksiimiroshnyk, line 698: I think it would be good to remove sku.getSkuMedia / setSkuMedia methods from Sku.
Also, from the product we can remove deprecated defaultParentCategoryId column...

oleksiimiroshnyk (commented):
guess deprecated fields/methods in Product/Sku can be removed with this PR

oleksiimiroshnyk (changes requested):
I think it would be good to remove sku.getSkuMedia / setSkuMedia methods from Sku.
Also, from the product we can remove deprecated defaultParentCategoryId column...

oleksiimiroshnyk (approved):

PR #2912, opened 2023-10-17
Improve hashCode method for the ProductImpl

troian88:
Added ID to the hashCode method

https://github.com/BroadleafCommerce/QA/issues/5055

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/catalog/domain/ProductImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/catalog/domain/ProductImpl.java
@@ -1081,7 +1081,7 @@ public class ProductImpl implements Product, ProductAdminPresentation, Status, A
 
     @Override
     public int hashCode() {
-        return Objects.hash(id, skus);
+        return Objects.hash(id, additionalSkus);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/catalog/domain/ProductImpl.java
```diff
@@ -1080,10 +1081,7 @@ public Map<String, Set<String>> getProductOptionValuesMap() {
 
     @Override
     public int hashCode() {
-        final int prime = 31;
-        int result = 1;
-        result = prime * result + ((skus == null) ? 0 : skus.hashCode());
-        return result;
+        return Objects.hash(id, skus);
```
oleksiimiroshnyk: I think the idea was not to use skus but additional skus and or(in case additional skus are null/empty)
As now skus collection is transient and filled in getSkus() method which is deprecated and used only in some places so chances that skus collection will be not initialized are pretty high. which leads to hascode being a constant... Including Id is better, but still I think we should not use skus collection
So better not rely on it and use additional/default sku...

oleksiimiroshnyk (changes requested):

core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/catalog/domain/ProductImpl.java
troian88: updated

oleksiimiroshnyk (approved):

PR #2923, opened 2023-10-27
Remove some deprecated classes

riteshadhikari17:
**A Brief Overview**
Removing some deprecated classes which are not used.

**Additional context**
Not exactly part of this [QA](https://github.com/BroadleafCommerce/QA/issues/4923) but related


common/src/main/java/org/broadleafcommerce/common/util/ResourceBundleExtensionPoint.java
KatrukOV, line 1: need remove this (line 72) * @param resourceBundleExtensionPoint

common/src/main/java/org/broadleafcommerce/common/util/BroadleafMergeResourceBundleMessageSource.java
```diff
@@ -31,7 +31,7 @@
 
 
 /**
- * <p>Replaces the deprecated {@link BLResourceBundleMessageSource} by using a {@link ReloadableResourceBundleMessageSource}
+ * <p>Replaces the deprecated BLResourceBundleMessageSource by using a {@link ReloadableResourceBundleMessageSource}
```
KatrukOV: Maybe it's better to remove this test?

KatrukOV (approved):

common/src/main/java/org/broadleafcommerce/common/util/BroadleafMergeResourceBundleMessageSource.java
riteshadhikari17: Which test? Could you explain what do u mean?

KatrukOV: *text

riteshadhikari17: Oh ok. I thought it would be useful info for someone if they were still using the old class. I can re-word it and say `Replaces the older BLResourceBundleMessageSource` instead.

PR #2944, opened 2023-12-05
Fix to better check if sequence needs to be updated for hibernate 6

oleksiimiroshnyk:
- Fix to better check if sequence needs to be updated for hibernate 6
In hibernate 6 sequence generators updated formulas, so seq corruption detector should not only increment correctly, but a condition to check if sequence needs to be update should be enhanced

Fixes: BroadleafCommerce/QA#5157


riteshadhikari17:
The PR build failed. Its weird error about VM
```
[ERROR] Error occurred in starting fork, check output in log
[ERROR] Process Exit Code: 1
[ERROR] org.apache.maven.surefire.booter.SurefireBooterForkException: The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
```

oleksiimiroshnyk:
Pushed another fix, think it should be better now, let's see

riteshadhikari17 (approved):

PR #2971, opened 2024-02-19
Fixes for save button/title bar UI issues when shrinking/resizing window and save button was not properly placed, and updates its placement after scroll event

oleksiimiroshnyk:
- Fixes for save button/title bar UI issues when shrinking/resizing window and save button was not properly placed, and updates its placement after scroll event
 
Accepted changes proposed by Flywire

Fixes: BroadleafCommerce/QA#5209

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/css/admin/blc-admin.css
+++ b/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/css/admin/blc-admin.css
@@ -718,12 +718,12 @@ body.login .account-actions ul li a {
 }
 
 .app-content .content .section-content .main-content .content-area-title-bar {
-    width: 100%;
+    width: 100vw;
     height: 80px;
     line-height: 80px;
     background-color: white;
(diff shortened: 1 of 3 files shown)


StanislavFedorov (approved):

riteshadhikari17:
Works fine when resizing but as the width gets very narrow view, things start overlapping. Is there better way to handle this? Can we define certain width after which it stops and goes off the view/window? Just like preview button on top.

https://github.com/BroadleafCommerce/BroadleafCommerce/assets/35038090/e9e7760a-a01b-4a27-b53b-f645e7ff9cf6

oleksiimiroshnyk:
> Is there better way to handle this?

Probably so, but it can take a lot of time with possibly page redesign - as original design wasn't made to handle this things so there is a chance that every little piece will need to be reviewed to have some min/max width when it can shrink or move. And given that this thing bothers flywire only and they are fine with the fix(as they mostly provided it) I see no reason in investing time in this..

riteshadhikari17 (approved):

PR #2975, opened 2024-03-13
Fixes the issue where it was showing incorrect placement of error message

KatrukOV:
**A Brief Overview**
in entityForm.css has been changed value float to 'none' of headerFlashAlertBoxContainer

**Link to QA issue**
[QA-5221](https://github.com/BroadleafCommerce/QA/issues/5221)

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/css/admin/entityForm.css
+++ b/admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/css/admin/entityForm.css
@@ -326,6 +326,12 @@ button:disabled:hover, button.disabled:hover {
  ADMIN HEADER FLASH RELATED STYLES
 ****************************************/
 #headerFlashAlertBoxContainer {
+    float: right;
+    padding-right: 10px;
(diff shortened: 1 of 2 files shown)


admin/broadleaf-open-admin-platform/src/main/resources/open_admin_style/css/admin/entityForm.css
```diff
@@ -326,7 +326,7 @@ button:disabled:hover, button.disabled:hover {
  ADMIN HEADER FLASH RELATED STYLES
 ****************************************/
 #headerFlashAlertBoxContainer {
-    float: right;
+    float: none;
```
oleksiimiroshnyk, line 335: I see tag id is used also in a 2 htmls, so this change will affect them as well.
I don't know if this will break anything for them, so I suggest create a new css class and assign it to that dynamically created div.
Or simply in the place of that div creation add a style attribute

oleksiimiroshnyk (changes requested):

oleksiimiroshnyk (approved):

PR #3012, opened 2024-07-04
Fix for the Error is triggered while attempting to edit items from Wishlist

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/5279

**A Brief Overview**
Fix for the Error is triggered while attempting to edit items from Wishlist


core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/order/service/OrderServiceImpl.java
```diff
@@ -889,6 +889,15 @@ protected Order removeItemInternal(Long orderId, Long orderItemId, boolean price
         return context.getSeedData().getOrder();
     }
 
+    protected Order findOrderByIdOrByOrderItemId(final Long orderId, final Long orderItemId) {
```
KatrukOV, line 892: Perhaps better "findOrderByIdOrOrderItemId"

KatrukOV (approved):

PR #3013, opened 2024-07-04
Ability to choose location for currency

KatrukOV:
**A Brief Overview**
Added currencyCodes to apply currency location logic.

Needed to set properties or will be used default value (for example):
currency.location=SUFFIX
currency.codes.location.apply=EUR,USD


riteshadhikari17:
@KatrukOV its looks like we merged other [PR](https://github.com/BroadleafCommerce/BroadleafCommerce/pull/3007). Is it ok to close this?

PR #3029, opened 2024-10-29
Antisamy updates properties

KatrukOV:
**A Brief Overview**
Antisamy common-regexps paragraph value was modified in library

**Additional context**
[https://github.com/BroadleafCommerce/QA/issues/5319](https://github.com/BroadleafCommerce/QA/issues/5319)


KatrukOV:
@Jerry77oz @danielcolgrove @oleksiimiroshnyk 
Maybe this file (xml) is better to remove from the framework and use it from an external library?

riteshadhikari17 (approved):

riteshadhikari17:
After discussion it was decided to keep the file to provide a baseline for security. Clients can always override it.

PR #3056, opened 2025-03-01
Added a property to enable a “guaranteed” inventory 

troian88:
**A Brief Overview**
Added lock and invalidation cache

**Additional context**
https://github.com/BroadleafCommerce/QA/issues/5357

### DIFF
--- a/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/inventory/service/InventoryServiceImpl.java
+++ b/core/broadleaf-framework/src/main/java/org/broadleafcommerce/core/inventory/service/InventoryServiceImpl.java
@@ -17,6 +17,7 @@
  */
 package org.broadleafcommerce.core.inventory.service;
 
+import jakarta.annotation.PostConstruct;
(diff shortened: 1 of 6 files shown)


danielcolgrove:
Next step here:
- Let's add an endpoint to decrement for one product
- Create a jmeter load test to exercise that endpoint
- This should give us a good test for the final solution.  You can even test this current solution to see if there is a gap.

danielcolgrove:
@troian88 A couple thoughts:
1) The most common issue will be a lock attempt failing to obtain a lock.  We want a decent retry windows/number of attempts.  Then, of course, have a unique exception (which you have) if it finally fails.
2) Another tolerance piece is being able to detect an orphaned lock and recover from it.  I do like the approach you are taking (interface, reentrant lock, etc) but it may be worth having a lock timeout saved to the database and checked.  It may require a dedicated lock monitor or something to prevent a lock timeout to suddenly have multiple threads unlocking and taking ownership.  Maybe that's too difficult to accomplish??

PR #3062, opened 2025-04-10
Remove cast from SQL where clause

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/5371

**A Brief Overview**
Remove cast from SQL where clause


danielcolgrove:
@StanislavFedorov was there any indication in the code history as to why the cast was added?  Perhaps when and by whom it was done?

StanislavFedorov:
> was there any indication in the code history as to why the cast was added? Perhaps when and by whom it was done?

There are no indications. This code has existed since 2015. Locally for me, it works fine.
This code anyway is translated to SQL, so for me this conversion looks redundant. But let's see the customer's feedback

riteshadhikari17 (approved):

PR #3065, opened 2025-05-08
Avoid setting readOnly flag if a transaction is not NEW

Jerry77oz:
Postgres emits a WARN when setting the `readOnly` flag on a `NEW` transaction:

```log
2025-02-20 13:46:19.975 ERROR (admin) SqlExceptionHelper - ERROR: cannot execute UPDATE in a read-only transaction
```

Add logic to prevent setting the flag if the transaction is detected to be `NEW`


danielcolgrove:
@Jerry77oz Is there a QA issue for this?  I didn't see it linked.

PR #3067, opened 2025-05-12
Upgrade tomcat and http2

riteshadhikari17:
Upgrade some libs for security.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,7 +56,7 @@
         <groovy.version>4.0.20</groovy.version>
         <gpg.plugin.version>3.1.0</gpg.plugin.version>
         <broadleaf-presentation.version>2.0.1-GA</broadleaf-presentation.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -1006,8 +1006,18 @@
...
+                        <groupId>org.eclipse.jetty.http2</groupId>
+                        <artifactId>http2-common</artifactId>
+                    </exclusion>
                 </exclusions>
             </dependency>
+            <!-- Override http2-client version coming from solr-->
+            <dependency>
+                <groupId>org.eclipse.jetty.http2</groupId>
+                <artifactId>http2-common</artifactId>
+                <version>11.0.25</version>
```
danielcolgrove, line 1019: Should we use a property value instead?

danielcolgrove (approved):

pom.xml
riteshadhikari17: I don't think it's necessary. This is how we usually do it for transitive dependency which we don't actively use in framework.

PR #3091, opened 2025-10-28
Add validation to prevent usage of special characters in offer targeters

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/5424

**A Brief Overview**
Add validation to prevent the usage of special characters in offer targeters

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/validation/SkuNamePropertyValidator.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/service/persistence/validation/SkuNamePropertyValidator.java
@@ -1,0 +1,77 @@
+/*-
+ * #%L
+ * BroadleafCommerce Open Admin Platform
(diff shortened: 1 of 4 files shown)


StanislavFedorov:
@riteshadhikari17 
Fixed tests

Jerry77oz:
@StanislavFedorov let's look at incorporating the latest [code/fix provided by Flywire](https://secure.helpscout.net/conversation/3085986081/57040):

Jerry77oz:
On more addition from the same thread:

## From the client
To align with the existing validator structure I believe we can add the
following check:

```java
protected String ERROR_KEY_CANNOT_END_WITH_COMMA = "skuNamePropertyValidatorCannotEndWithCommaError";

if (value.endsWith(",")) {
    return new PropertyValidationResult(false, ERROR_KEY_CANNOT_END_WITH_COMMA);
}
```

StanislavFedorov:
@Jerry77oz 
I have committed the last variant of the validation that we agreed with Mihai in the Slack channel

Jerry77oz (approved):

PR #3093, opened 2025-11-20
Fix for the issue Site customer does not login automatically after successful registration

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/5434

**A Brief Overview**
Fix for the issue Site customer does not login automatically after successful registration


danielcolgrove:
Tested and verified

danielcolgrove (approved):

PR #3099, opened 2026-01-15
Remove commons-collections from the framework in favor of commons-collections4

riteshadhikari17:
Commons-collections has security vulnerabilities which will not be fixed. So updating all imports to use `collections4` however there is some work around required for Commons BeanUtils library. The class `BeanToPropertyValueTransformer`  is hard-coded to depend on commons-collections.

### DIFF
--- a/admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/persistence/validation/OfferQualifyingCriteriaValidator.java
+++ b/admin/broadleaf-admin-module/src/main/java/org/broadleafcommerce/admin/persistence/validation/OfferQualifyingCriteriaValidator.java
@@ -17,7 +17,7 @@
  */
 package org.broadleafcommerce.admin.persistence.validation;
 
-import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.collections4.CollectionUtils;
 import org.broadleafcommerce.openadmin.dto.BasicFieldMetadata;
 import org.broadleafcommerce.openadmin.dto.Entity;
 import org.broadleafcommerce.openadmin.dto.FieldMetadata;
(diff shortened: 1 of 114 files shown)


riteshadhikari17:
@Jerry77oz / @StanislavFedorov / @troian88 This PR currently fails to compile because`BeanToPropertyValueTransformer` is used in the method of `filterProducts()` of the class `SearchFilterUtil` I am not comfortable messing with the code. Could one of you take over and find the work-around?

StanislavFedorov:
@riteshadhikari17 @Jerry77oz @danielBroadleaf 
I updated the code, and now the tests are fine.
But there is one important note.
The issue was in the **common-beanutils** component that is coupled with **commons-collections**.
**Commons Collections 4** is compatible only with **Commons BeanUtils** from version **2.0.0**.
Because older versions of commons-beanutils have a transitive dependency on commons-collections version 3.x.
I updated the common-beanutils to common-beanutils2, but they have only **M1** and **M2** versions of this component.
Could we use such releases in the Framework?
If not, then we need to rollback Commons Collections upgrade.

riteshadhikari17:
I am ok with using M2 of beanUtils. I can check periodically for official release and update when it comes out.

Jerry77oz:
I agree with @riteshadhikari17, we can use the M2. I'd rather do that then roll back to an unsecure library.

StanislavFedorov:
@riteshadhikari17 
>I am ok with using M2 of beanUtils. I can check periodically for official release and update when it comes out.

https://mvnrepository.com/artifact/org.apache.commons/commons-beanutils2

riteshadhikari17:
@Jerry77oz Could you approve the PR then?

Jerry77oz (approved):

PR #3116, opened 2026-04-09
Use BroadleafIdGenerator instead of deprecated GenericGenerator

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/5447

**A Brief Overview**
Use BroadleafIdGenerator instead of deprecated GenericGenerator

### DIFF
--- a/common/src/main/java/org/broadleafcommerce/common/persistence/IdOverrideTableGenerator.java
+++ b/common/src/main/java/org/broadleafcommerce/common/persistence/IdOverrideTableGenerator.java
@@ -104,13 +104,17 @@ public class IdOverrideTableGenerator extends TableGenerator implements Annotati
     @Override
     public void configure(GeneratorCreationContext creationContext, Properties params) throws MappingException {
         if (config != null) {
-            params.putIfAbsent(SEGMENT_VALUE_PARAM, config.segment_value());
-            params.putIfAbsent(ENTITY_NAME_PARAM, config.entity_name());
+            params.putIfAbsent(SEGMENT_VALUE_PARAM, config.segmentValue());
+            params.putIfAbsent(ENTITY_NAME_PARAM, config.entityName());
+            params.putIfAbsent(TABLE_PARAM, config.tableName());
+            params.putIfAbsent(SEGMENT_COLUMN_PARAM, config.segmentColumnName());
+            params.putIfAbsent(VALUE_COLUMN_PARAM, config.valueColumnName());
+            params.putIfAbsent(INCREMENT_PARAM, config.incrementSize());
         }
-        params.putIfAbsent("table_name", "SEQUENCE_GENERATOR");
-        params.putIfAbsent("segment_column_name", DEFAULT_SEGMENT_COLUMN_NAME);
-        params.putIfAbsent("value_column_name", DEFAULT_VALUE_COLUMN_NAME);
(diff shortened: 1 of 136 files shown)


common/src/main/java/org/broadleafcommerce/common/persistence/IdOverrideTableGenerator.java
```diff
@@ -104,8 +104,12 @@ public Serializable generate(final SharedSessionContractImplementor session, fin
     @Override
     public void configure(GeneratorCreationContext creationContext, Properties params) throws MappingException {
         if (config != null) {
-            params.putIfAbsent(SEGMENT_VALUE_PARAM, config.segment_value());
-            params.putIfAbsent(ENTITY_NAME_PARAM, config.entity_name());
+            params.putIfAbsent(SEGMENT_VALUE_PARAM, config.segmentValue());
```
danielcolgrove, line 107: Where is this constant defined?  Would it make sense to change all the other "static" names to be constants as well?

common/src/main/java/org/broadleafcommerce/common/persistence/IdOverrideTableGenerator.java
```diff
@@ -104,8 +104,12 @@ public Serializable generate(final SharedSessionContractImplementor session, fin
...
+            params.putIfAbsent(ENTITY_NAME_PARAM, config.entityName());
+            params.putIfAbsent("table_name", config.tableName());
+            params.putIfAbsent("segment_column_name", config.segmentColumnName());
+            params.putIfAbsent("value_column_name", config.valueColumnName());
+            params.putIfAbsent("increment_size", config.incrementSize());
```
danielcolgrove: I assume that with this new BroadleafIdGenerator we can still override the increment_size?  That is a common override for clients.

danielcolgrove (changes requested):

common/src/main/java/org/broadleafcommerce/common/persistence/IdOverrideTableGenerator.java
StanislavFedorov: Yep, these constants were defined in the Hibernate TableGenerator, changed

common/src/main/java/org/broadleafcommerce/common/persistence/IdOverrideTableGenerator.java
StanislavFedorov: We can set this value in the annotation, for example:
```
    @BroadleafIdGenerator(
            segmentValue = "ProductOptionImpl",
            entityName = "org.broadleafcommerce.core.catalog.domain.ProductOptionImpl",
            incrementSize = 100
    )
```

danielcolgrove: We can set it in the framework source code but can it be overridden by the client through config or extensions?

StanislavFedorov: hmm. I have not faced such customer customizations.
How can it be overridden in the current generator?

danielcolgrove: For these we can probably have then use java instrumentaton/class merging to override the annotations.

riteshadhikari17 (approved):

PR #3133, opened 2026-06-24
Do not allow blank admin user password (#5478)

StanislavFedorov:
https://github.com/BroadleafCommerce/QA/issues/5478

**A Brief Overview**
Do not allow a blank admin user password
Disabled by default, use `validate.admin.user.password` property to enable
@riteshadhikari17 
Please mention above in the release notes

### DIFF
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/config/OpenAdminSecurityConfig.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/config/OpenAdminSecurityConfig.java
@@ -17,6 +17,7 @@
  */
 package org.broadleafcommerce.openadmin.config;
 
+import org.broadleafcommerce.openadmin.security.BroadleafDaoAuthenticationProvider;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.context.annotation.Bean;
@@ -48,7 +49,7 @@ public class OpenAdminSecurityConfig {
 
     @Bean
     public AuthenticationProvider blAdminAuthenticationProvider() {
-        DaoAuthenticationProvider provider = new DaoAuthenticationProvider();
+        DaoAuthenticationProvider provider = new BroadleafDaoAuthenticationProvider();
         provider.setUserDetailsService(userDetailsService);
         provider.setPasswordEncoder(adminPasswordEncoder);
         return provider;
--- a/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/security/BroadleafDaoAuthenticationProvider.java
+++ b/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/security/BroadleafDaoAuthenticationProvider.java
@@ -1,0 +1,27 @@
+package org.broadleafcommerce.openadmin.security;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.security.authentication.BadCredentialsException;
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
+import org.springframework.security.authentication.dao.DaoAuthenticationProvider;
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.core.userdetails.UserDetails;
+
+public class BroadleafDaoAuthenticationProvider extends DaoAuthenticationProvider {
+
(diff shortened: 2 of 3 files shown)


danielcolgrove:
Do we have anywhere in the code where we establish minimum password requirements?  Perhaps we don't but we should.  For example, minimum 8, mixed case, one special character, etc.

StanislavFedorov:
@danielcolgrove 
We don't have any requirements for the admin user password.
Should we include it in this PR?

danielcolgrove:
I think we should.  Would it make sense to define a validation bean that can be overridden by the client?  Want to be able to allow them to override the rules based on their requirements.  Maybe input is password and output is a list of errors??  
badLength, badCase, badSpecialChar  (or something like that).  The basic view can just check if the list is not null and report invalid password.  Customers can get fancier and show individual errors.  Just thinking/typing out loud.

riteshadhikari17:
I am pretty sure we have something in the framework for validating password. Its set by`admin.password.regex.validation` property in [AdminUserImpl]( https://github.com/BroadleafCommerce/BroadleafCommerce/blob/8645873661b34fe0954cebe382aba59336714db0/admin/broadleaf-open-admin-platform/src/main/java/org/broadleafcommerce/openadmin/server/security/domain/AdminUserImpl.java#L118). But [the regex is simple](https://github.com/BroadleafCommerce/BroadleafCommerce/blob/8645873661b34fe0954cebe382aba59336714db0/admin/broadleaf-open-admin-platform/src/main/resources/config/bc/admin/common.properties#L59-L60)

StanislavFedorov:
@riteshadhikari17 
You are correct. I was confused because it was added relatively recently, and the requirements are weak.
But this PR, anyway, makes sense because 
 - blAdminRegexValidator doesn't mark the password field as required
- blAdminRegexValidator does not meet this requirement: `When a user tries to log in - present an error message stating password is required when the field is left blank.`

riteshadhikari17:
@StanislavFedorov These are few things suggested by gemini review.  Could you go over these?

**Duplication Concern:** You have implemented validatePassword inside AdminUserCustomPersistenceHandler and additionalAuthenticationChecks in BroadleafDaoAuthenticationProvider.

Recommendation: If the password validation logic grows more complex (e.g., regex requirements, history checks), these two places will get out of sync. Consider extracting this logic into a dedicated PasswordValidator service that both the Handler and the Provider can inject.

**The inspect Method:** You added an override for canHandleInspect and inspect. Ensure that by adding this, you aren't inadvertently exposing sensitive DTO fields to the UI that shouldn't be there, or causing overhead on the admin grid load.

**Logging:** In BroadleafDaoAuthenticationProvider.java, the log level is set to debug. Given this is a security validation failure, consider using info or warn level so that security audits can easily spot failed attempts in production logs without needing debug enabled globally.

Unit Tests: Have you added unit tests for BroadleafDaoAuthenticationProvider? Specifically, verify that it throws BadCredentialsException when the property is enabled and a blank password is provided.

Integration Tests: Verify that the AdminUser can still be saved via the Admin UI when validate.admin.user.password is false (to ensure no regressions).

StanislavFedorov:
@riteshadhikari17 
`Recommendation: If the password validation logic grows more complex (e.g., regex requirements, history checks), these two places will get out of sync. Consider extracting this logic into a dedicated PasswordValidator service that both the Handler and the Provider can inject.`

For complex Regex validation, we have blAdminRegexValidator. AdminUserCustomPersistenceHandler covers only the simple but critical case with a blank password

`You added an override for canHandleInspect and inspect. Ensure that by adding this, you aren't inadvertently exposing sensitive DTO fields to the UI that shouldn't be there, or causing overhead on the admin grid load.`

My changes only add a "required" flag for the password and confirmPassword fields, there are no other changes related to field visibility

`In BroadleafDaoAuthenticationProvider.java, the log level is set to debug. Given this is a security validation failure, consider using info or warn level so that security audits can easily spot failed attempts in production logs without needing debug enabled globally.`

Changed to **warn**

riteshadhikari17:
@Jerry77oz / @danielcolgrove  Could one of you review and merge this?

danielcolgrove:
The only other concern is that there is a MultiTenantAdminUserCustomPersistenceHandler.  That handler does not call to super so the field property changes would get missed.

StanislavFedorov:
>The only other concern is that there is a MultiTenantAdminUserCustomPersistenceHandler. That handler does not call to super so the field property changes would get missed.

Thanks for this note! MultiTenantAdminUserCustomPersistenceHandler handles other UI - it works when we create an Admin user from the Site section.
But we need to apply this logic here too

